Issue #22, opened 2017-05-11
new repo to CCPPETMR-SuperBuild


bathomas:

KrisThielemans:
As discussed in our 14th software meeting. Motivation: the SuperBuild builds more than SIRF (gadgetron, ismrmrd-to-siemens etc).

Note that this updating of Wiki and other documentation

KrisThielemans:
too dangerous to do this now. We'll postpone to after 0.9.0 release.

KrisThielemans:
@casperdcl can we discuss how this fits in with your unification strategy? Name for the unified project might not be `*SuperBuild` anyway.

casperdcl:
we can rename this to whatever we want, not sure what that has to do with unification...

1. CCPPETMR/SIRF
2. CCPPETMR/<whatever we call this>
3. CCPPETMR/VM (ideally to be merged into 2)

KrisThielemans:
move CCPPETMR_VM files into this repo. move its wiki files to SIRF wiki.

casperdcl:
Think we can rename it to just `SuperBuild`. The way redirection works with github means previous clones should continue to work.

PR #34, opened 2017-09-14
Docker

casperdcl:
alternative to https://github.com/CCPPETMR/SIRF/issues/29


casperdcl:
not yet finished with this

casperdcl:
OK this works on linux out-of-the-box. docker pulls and builds an image which takes up about 

On windows (as noted [in the wiki](https://github.com/CCPPETMR/SIRF/wiki/SIRF-SuperBuild-on-Docker)) you'd have to install a desktop / vnc server within the container and connect for a GUI. Will add docker files for that later...

Dockerfile
```diff
@@ -92,7 +82,19 @@ RUN cd SIRF-SuperBuild \
   && make -j8
```
bathomas: - ```make -j``` rather than ```make -j8``` ?
- With the UID/GIDs, is the plan to use something like ```--build-arg UID=$(id -u)``` during building? If so, we'd probably need to document this somewhere.
- Add ```--no-install-recommends``` and delete apt-get cache. Should save a bit of disk space.

casperdcl:
@bathomas 

- removed `8` from `make -j`
- [wiki](https://github.com/CCPPETMR/SIRF/wiki/SIRF-SuperBuild-on-Docker) updated with `--build-arg` notes
- `--no-install-recommends` added, possibly reduces image size by <1% but I'll leave it there anyway

casperdcl:
Note that an image based on `alpine` might be relatively painless to create and would be roughly half the size but also half the speed when running, so I don't think it's worth it.

KrisThielemans:
There is obviously overlap with CCPPETMR_VM. Most obvious is https://github.com/CCPPETMR/CCPPETMR_VM/blob/master/scripts/INSTALL_prerequisites_with_apt-get.sh, but also some other content of its Vagrantfile.

It makes me wonder if the `Dockerfile` belongs in https://github.com/CCPPETMR/CCPPETMR_VM, or at least should use some of it.

For discussion tomorrow (if there's time)

casperdcl:
I feel we should really have the VM repo deprecated and vagrantfiles/dockerfiles etc put here. I think we raised this point before but not sure what the conclusion was.

KrisThielemans:
it's a possibility of course. I don't have a strong opinion either way. Yet another discussion point!

bathomas:
I agree with @casperdcl that ```alpine``` images are not really worth it yet. Perhaps something to keep in mind for the future. I recently built ITK in an ```alpine``` [container](https://hub.docker.com/r/benthomas1984/alpine-itk/) and it is considerably smaller than with ```ubuntu```. In theory I suppose we could use Docker hub's automated building, but in reality I suspect the build would not complete within the current 2 hour/1 core limit.

casperdcl:
Yes...

To get around issues with installing gadgetron in windows (that's still an issue, right?) I was also thinking of splitting gadgetron into its own container service and connecting to it. That would mean Windows users could install everything locally and just have gadgetron in a container.

KrisThielemans:
windows: yes, that's what we currently do, but via a VM, not Docker. Gadgetron does supply a docker image, so for people having docker on Windows (not trivial as dicussed before) that'd work as well. Probably worth documenting, if someone ever tries this out.

docker/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+  build-essential      \
+  git                  \
+  g++                  \
+  man                  \
+  make                 \
+  sudo
+
+# CMake
+RUN mkdir /opt/cmake
+RUN curl https://cmake.org/files/v3.7/cmake-3.7.2-Linux-x86_64.sh > cmake.sh
```
KrisThielemans: need to install curl?

docker/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+RUN curl -O https://raw.githubusercontent.com/git/git/master/contrib/completion/git-prompt.sh
+
+# Python (virtualenv)
+RUN curl -O https://bootstrap.pypa.io/get-pip.py
+RUN sudo -H python get-pip.py
+RUN rm get-pip.py
+RUN sudo -H python -m pip install -U pip virtualenv
+RUN virtualenv py2
+ARG PIPINST="$HOME/py2/bin/pip install -U"
+RUN $PIPINST cython
```
KrisThielemans: cython not used?

docker/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+
+RUN git clone https://github.com/CCPPETMR/SIRF-SuperBuild
+WORKDIR $HOME/SIRF-SuperBuild
+RUN . ~/py2/bin/activate \
+  && /opt/cmake/bin/cmake .
+RUN make -j
+
+RUN mv INSTALL/share/gadgetron/config/gadgetron.xml.example INSTALL/share/gadgetron/config/gadgetron.xml
```
KrisThielemans: I guess will no longer be necessary after accepting Edo's PR. possibly test for existence first.

docker/devel/shared.txt
```diff
@@ -0,0 +1,12 @@
+This file's directory and all its contents are mounted as follows:
+
+Host:
+
+`/path/to/SIRF-Superbuild/devel`
```
KrisThielemans: not sure where this comes from, but as we'll rename the repo ( #22) , this would then have to be changed as well?

docker/Dockerfile
casperdcl: I did install `curl` at the top of the Dockerfile as I use it often to download files, and it's a small install.

docker/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+
+# git completion
+RUN curl -O https://raw.githubusercontent.com/git/git/master/contrib/completion/git-prompt.sh
+
+# Python (virtualenv)
+ADD user_python-ubuntu.sh .
+RUN bash user_python-ubuntu.sh
+RUN rm user_python-ubuntu.sh
+
+RUN git clone https://github.com/CCPPETMR/SIRF-SuperBuild
```
KrisThielemans: let's put the actual build stuff also in its own script. Then what is here is just Docker-env setup, and calling a few scripts. I think that's nicest.

docker/Dockerfile
```diff
@@ -0,0 +1,77 @@
+FROM ubuntu:16.04
+MAINTAINER Casper da Costa-Luis <imaging@caspersci.uk.to>
+
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update && apt-get upgrade -y --no-install-recommends
+
+# Set locale, suppress warnings
+RUN apt-get update && apt-get install -y apt-utils curl locales
```
KrisThielemans: I hadn't seen the curl here. Any reason not to put it in build-essentials? That would make sense as the scripts depend on it, so best to make sure it's there (then the scripts are self-contained)

docker/user_sirf-ubuntu.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# git completion
+curl -O https://raw.githubusercontent.com/git/git/master/contrib/completion/git-prompt.sh
```
KrisThielemans: sorry. this doesn't seem to belong here (if you'd imagine this script to be run outside of the docker set-up). I think it's better in the Dockerfile.

docker/user_sirf-ubuntu.sh
```diff
@@ -0,0 +1,12 @@
...
+
+git clone https://github.com/CCPPETMR/SIRF-SuperBuild
+pushd SIRF-SuperBuild
+source ~/py2/bin/activate && /opt/cmake/bin/cmake .
```
KrisThielemans: I think the `+source ~/py2/bin/activate` should be outside when calling this script. Otherwise it introduces a dependency again.

docker/user_sirf-ubuntu.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# git completion
+curl -O https://raw.githubusercontent.com/git/git/master/contrib/completion/git-prompt.sh
+
+git clone https://github.com/CCPPETMR/SIRF-SuperBuild
+pushd SIRF-SuperBuild
```
KrisThielemans: cd?

docker/user_sirf-ubuntu.sh
casperdcl: moved to .bashrc

docker/user_sirf-ubuntu.sh
casperdcl: made optional now.


the problem is that means the caller must do this on the same line (`&&`) or must `source` in bash. Dockerfiles can't do the latter afaik.

casperdcl:
@KrisThielemans yes we could move then to a scripts subdir but 1) if that's in a different repo then yes, would have to install `git` and clone first. 2) would have to move scripts to a subdir of the docker source in order to be visible to the Dockerfile.

Do you think we should rely less on the SuperBuild and actually `USE_SYSTEM_*` more (like in `.travis.yml`)? This would reduce build time and image size, but might mean more pain for us maintainers to update.

KrisThielemans:
I think `USE_SYSTEM_*` is fine for inside Docker or Virtualbox, as we have control over the OS and packages (ok, not really if there's an `apt-get update` in there). We had this discussion for Vagrant as well. @bathomas, opinions?

docker/user_sirf-ubuntu.sh
KrisThielemans: are you saying that in the Dockerfile, you cannot do something like
```
RUN source ~/py2/bin/activate && user_sirf-ubuntu.sh
```

docker/user_sirf-ubuntu.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+export PATH=/opt/cmake/bin:"$PATH"
```
KrisThielemans: similar dependency on docer environment as with the `activate`. Maybe we cannot get round this.

docker/user_sirf-ubuntu.sh
casperdcl: Yes, this doesn't work (subsequent RUN commands will not inherit the `py2` environment)

docker/user_sirf-ubuntu.sh
casperdcl: Should be fine though as

1. this path prefix won't survive in the built image (well, it will, but only thanks to .bashrc)
2. if this path doesn't exist it'll be ignored (e.g. if the script is called by someone with `cmake` in their PATH elsewhere

KrisThielemans: ok. maybe put it at the end? (I guess in Docker you 'll know there won't be another one)

casperdcl: removed

docker/user_sirf-ubuntu.sh
casperdcl: ah yes. just need to add .bashrc before this step, done now

casperdcl:
So I suppose all issues here are now sorted apart from `USE_SYSTEM_*`?

KrisThielemans:
yes. looks fine to me. We can do the `USE_SYSTEM_*` later. Squash-merge?

casperdcl:
Yes, I've been at the summer school thing so been contributing via terminal on phone. Hard to test things... Will `--no-ff` merge, as it's better to track the feature branch history without making it hard to see `master` history. Already rebased to remove superfluous commits.

PR #36, opened 2017-09-20
flexible version tagging

paskino:
Enables flexible version tagging for Gadgetron, SIRF, STIR and ISMRMRD.

Devel build (1/0) : latest master from upstream, stable SIRF 0.9.0 versions
Additionally one may specify which URL/TAG to use for any of the sub-projects.


paskino:
`OPTION`s are boolean so they cannot be used for what you want.

KrisThielemans:
Travis worked (which is great news as it means we're still ok with Gadgetron and STIR master).

I did notice that setting `DEVEL_BUILD=On` after the first configure doesn't reset the `GIT_TAG`s. That seems unavoidable (but might mean that `DEVEL_BUILD` shouldn't be a cached variable. Possibly @bathomas or @casperdcl has some ideas on this.

Finally, I think it would be good to prefix all these advanced variables with the same thing. That makes it easier to find them (e.g. when displaying as "Grouped" in CMake. Maybe `SB_SIRF_TAG` (for "SuperBuild") or so?

paskino:
I've added the build with `DEV_BUILD=ON` to travis and it works.

paskino:
I've noticed that the advanced variables get grouped anyway by package name: SIRF, ISMRMRD, Gadgetron  and STIR, which is maybe even better.

PR #45, opened 2017-09-27
Use Matlab_ROOT_DIR as opposed to MATLAB_ROOT

KrisThielemans:
WARNING: This breaks backwards compatibility.

Matlab_ROOT_DIR is the variable used by CMake's FindMatlab.cmake so is what we should use as well.

Currently we use this to set both variables for SIRF (as SIRF 0.9.0 used MATLAB_ROOT).


KrisThielemans:
If you want to try SIRF FindMatlab branch, use
```
cmake -DSIRF_TAG=FindMatlab 
```
Potentially add `-DMatlab_ROOT_DIR=....` if you need it (and for testing)

KrisThielemans:
merging this will need updates to the wiki.

PR #58, opened 2017-11-18
ccache within CMake

KrisThielemans:
according to what I've read, this should have speeded-up a rebuild on Travis (at least on Linux), but timing was virtually the same. I haven't tried this locally yet.

I wonder if it doesn't work because because `language` is `Python` in `.travis.yml`.

See also https://docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 # Language selection
 language: python
 
+cache: ccache
+
 # Currently commented out as set with environment flags
 # Both clang and gcc can be tested. The more is the better.
 #compiler:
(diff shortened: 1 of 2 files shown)


KrisThielemans:
We need to indeed remove `language: python` (as done elsewhere). However, this actually doesn't speed-up local compiles either. The reason is that the CMake setting to use `ccache` 
```
set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
```
doesn't get transported to the "external" projects.

It's possible that the symbolic linking trick for `ccache` is the only work around (aside from changing the CMakeLists.txt for every project of course)

casperdcl:
so this currently works on travis (using the symlink method) but this issue is looking for better cmake support

KrisThielemans:
Won't fix this as it seems up to the depend projects to take care of `ccache`. Developers will have to use the linking trick if they're interested in `ccache` for a project that doesn't have support for it in its CMake files.

Issue #49, opened 2017-10-10
overlap between EXTERNAL_PROJECT_BUILD_TYPE and CMAKE_BUILD_TYPE 


KrisThielemans:
these 2 are the same in our case so remove the former if possible

KrisThielemans:
@bathomas check if `EXTERNAL_PROJECT_BUILD_TYPE` is exposedto the user as cached variable. if so, remove it from the cache. Set it from `CMAKE_BUILD_TYPE`

KrisThielemans:
closed via #65

PR #65, opened 2018-01-05
Align EXTERNAL_PROJECT_BUILD_TYPE with CMAKE_BUILD_TYPE

bathomas:
Remove `EXTERNAL_PROJECT_BUILD_TYPE` from CMake GUI and make it equal to the value of `CMAKE_BUILD_TYPE`. #49.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -45,9 +45,8 @@ set (SUPERBUILD_INSTALL_DIR ${CMAKE_INSTALL_PREFIX})
 
 include(ExternalProject)
 
-set(EXTERNAL_PROJECT_BUILD_TYPE "Release" CACHE STRING "Default build type for support libraries")
-set_property(CACHE EXTERNAL_PROJECT_BUILD_TYPE PROPERTY
-STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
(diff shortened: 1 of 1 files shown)


SuperBuild.cmake
```diff
@@ -45,9 +45,8 @@ set (SUPERBUILD_INSTALL_DIR ${CMAKE_INSTALL_PREFIX})
 
 include(ExternalProject)
 
-set(EXTERNAL_PROJECT_BUILD_TYPE "Release" CACHE STRING "Default build type for support libraries")
-set_property(CACHE EXTERNAL_PROJECT_BUILD_TYPE PROPERTY
-STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+set(EXTERNAL_PROJECT_BUILD_TYPE "${CMAKE_BUILD_TYPE}" CACHE INTERNAL "Default build type for support libraries")
```
KrisThielemans: isn't using a cached variable potentially confusing? Say I run CMake once, then change `CMAKE_BUILD_TYPE`, won't `EXTERNAL_PROJECT_BUILD_TYPE` keep its first value?

bathomas: Caching as an `internal` variable keeps `CMAKE_BUILD_TYPE` and `EXTERNAL_PROJECT_BUILD_TYPE` in sync with one another. Also has the benefit of hiding it from the user in the UI.

PR #68, opened 2018-01-12
Itk

rijobro:
Optionally build STIR with ITK


SuperBuild/External_ITK.cmake
```diff
@@ -0,0 +1,80 @@
...
+        -DCMAKE_INSTALL_PREFIX=${${proj}_Install_Dir}
+	      -DBUILD_SHARED_LIBS=true
+        -DCMAKE_BUILD_TYPE=release
+        -DBUILD_TESTING=false
+        -DBUILD_EXAMPLES=false
+	  INSTALL_DIR ${${proj}_Install_Dir}
+    DEPENDS ${${proj}_DEPENDENCIES}
+  )
+
+  set(${proj}_ROOT        ${${proj}_SOURCE_DIR})
```
KrisThielemans: `ITK_ROOT` is not used. The variable for ITK is `ITK_DIR` as it uses ` ITKConfig.cmake` (as opposed to a  `FindITK.cmake` module). I wouldn't really know how to set `ITK_DIR` though. It's likely going to be OS dependent. Possibly @bathomas knows how to do this. But luckily it seems we don't need this (as it seems to be found because of `CMAKE_INSTALL_PREFIX`).

See also #70 which contains a link to the Slicer SuperBuild file for ITK.

SuperBuild/External_STIR.cmake
```diff
@@ -58,6 +57,28 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
         -DDISABLE_ITK=On
         # Use 2 variables for ROOT to cover multiple STIR versions
         -DDISABLE_CERN_ROOT_SUPPORT=ON -DDISABLE_CERN_ROOT=ON
+   )
+
+  # Append CMAKE_ARGS for ITK choices
+  if (USE_ITK AND USE_SYSTEM_ITK)
+    set(STIR_CMAKE_ARGS 
+      ${STIR_CMAKE_ARGS} 
+      -DDISABLE_ITK=false
```
KrisThielemans: remove this line. STIR will by default look for it, so no need to disable the disable...

SuperBuild/External_STIR.cmake
```diff
@@ -58,6 +57,28 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
...
+        # If USE_SYSTEM_ITK, ITK_DIR will have been set via find_package in External_ITK.cmake
+        # If !USE_SYSTEM_ITK, ITK_DIR will get set during the installation of ITK
+      -DITK_DIR=${ITK_DIR}
+    )
+  # ITK not required, so disable
```
KrisThielemans: I guess move the comment inside the `else()`

version_config.cmake
```diff
@@ -73,6 +73,10 @@ option(DEVEL_BUILD "Use current versions of major packages" OFF)
 set(googletest_URL https://github.com/google/googletest )
 set(googletest_TAG release-1.8.0)
 
+## ITK
+set(ITK_URL git://itk.org/ITK.git)
```
KrisThielemans: we're using `https` links for all the others. Not 100% sure what the impact of using `git` is. best to be uniform I guess.

Issue #50, opened 2017-10-10
add USE_SYSTEM_GTest variable


KrisThielemans:
Do this after fixing #19

paskino:
Closed with https://github.com/CCPPETMR/SIRF-SuperBuild/commit/27e3580664a2408aa8941fee3e0ccdc5f3c5390c

PR #69, opened 2018-01-12
Uses GTest rather than googletest

### LINKED ISSUE #19: rename googletest to GTest
CMake uses `find_package(GTest)`. Best to stick with that

paskino:
Closes #19 and #50 
`find_package(GTest)` did not find the GTest on my VM, though


KrisThielemans:
`find_package(GTest)` indeed doesn't find GTest on Ubuntu as it's a source-only package. Gadgetron knows about this and then [finds the source and builds it](https://github.com/gadgetron/gadgetron/blob/f51d0d04a19e89b4526ed82e3aed0b078602e929/CMakeLists.txt#L226). I don't understand enough about GTest, but this looks fine to me.

PR #86, opened 2018-02-08
Travis, docker, osx improvements

casperdcl:
- [x] make travis build docker images
  + [x] auto push to registry ([hub.docker.com/r/ccppetmr/sirf](https://hub.docker.com/r/ccppetmr/sirf/):latest)
    * [x] separate :devel tag
    * [x] push git tag if defined (:release)
    * [x] cache docker layers to avoid excessive rebuilds and registry up/downloads
      - ~~`docker save` and `load` on travis?~~
      - related to https://github.com/docker/compose/issues/883 -> use `docker-compose` or `docker`, not both
  + [x] documentation/README
- [x] fix errors in `if` blocks
  + [x] stop silently ignoring errors
  + [x] fix `brew tap homebrew/science` deprecation ([Homebrew/homebrew-science](https://github.com/Homebrew/homebrew-science))
- [x] use entrypoint & gosu for better solution to build-arg U/GID on linux?
  + [x] make codacy happy
  + https://github.com/moby/moby/issues/7198
  + https://github.com/boxboat/fixuid
  + https://denibertovic.com/posts/handling-permissions-with-docker-volumes/


casperdcl:
@rijobro hopefully you're fine with the removal of `brew tap homebrew/science`

docker/Dockerfile
```diff
@@ -10,73 +10,24 @@ LABEL \
...
 # SIRF
+COPY devel/.ccache/ ccache/
+RUN sudo chown -R $(id -un) ccache/
+RUN sudo chgrp -R $(id -gn) ccache/
+RUN ccache -o cache_dir="$PWD"/ccache
 COPY user_sirf-ubuntu.sh .
 RUN EXTRA_BUILD_FLAGS="-DCMAKE_BUILD_TYPE=Release -DUSE_SYSTEM_Boost=ON \
-  -DUSE_SYSTEM_SWIG=ON -DUSE_SYSTEM_HDF5=ON -DUSE_SYSTEM_FFTW3=ON \
-  -DUSE_SYSTEM_Armadillo=ON" \
+  -DDEVEL_BUILD=ON -DUSE_SYSTEM_Armadillo=ON -DUSE_SYSTEM_FFTW3=ON \
```
KrisThielemans: seems to say our docker images are built with most recent masters. Good idea?

casperdcl: yes:

1. all travis builds will trigger a docker build
    a) the docker build will, however, only use master branches (`DEVEL_BUILD`)
    b) the built image will be tagged `latest` and pushed to hub.docker.com only if tests pass in the container

arguments for (1):
- though the source code may not change, external dependencies (`ubuntu:16.04`, `apt-get` packages, etc) may have changed so better to rebuild
- everything's cached (both docker layers and ccache within docker) so this is quick and incremental - it's actually quicker than mac osx builds!

arguments for (1a):
- `master` should always be stable

arguments for (1b):
- one less thing to do when releasing (don't have to build or have docker locally)
- when we want to make a new release, we simply locally tag the current `latest` image as also being `v1.0.0` etc so that it never gets overwritten by future builds
    + might want to automate this in .travis.yml

arguments against anything:
- ?

casperdcl: I also don't intend to sync the docker-specific bits of this with the `.travis.yml` in `SIRF` since that would seem excessive in terms of number of updates to hub.docker.com/r/cccppetmr/sirf ... And the important thing is releases, ie. version_config.cmake, which is not in `SIRF`

KrisThielemans: not quite sure about this really. It means th edocker image will always be made with the very latest Gadgetron etc. We have no clue what they are doing, so it might easily break something. I would guess that we could work with SIRF/master only. Of course, that might require a more recent Gadgetron/STIR than what's currently the default in version_config.cmake. hmmm.

casperdcl: ok, now changed. docker/ccppetmr/sirf:latest doesn't have DEVEL_BUILD, while docker/ccppetmr/sirf:devel does

PR #87, opened 2018-02-12
Consistent build dir

### LINKED ISSUE #16: make location of output of SuperBuild consistent
some projects (probably CMake-ones) have their build in `x-prefix/src/x-build` as opposed to `x-build`.

Also, INSTALL is currently hard-wired but should honour `CMAKE_INSTALL_PREFIX`.

paskino:
Creates a consistent build directory according to [this specifications](https://github.com/CCPPETMR/SIRF-SuperBuild/issues/16#issuecomment-360772097):

- `sources`
    - `SIRF`
    - `STIR`
- ` builds`
    - `SIRF`
       - ` build`
       - ` stamp`
       - `tmp`
    - `STIR`
- `install`
 
Adds also the option `BUILD_OPENMP` but is currently not used.

Builds fine and tests run and checked.

Closes #16


KrisThielemans:
could you rename `BUILD_OPENMP` to something like `STIR_USE_OPENMP` (check best naming)? It'll be the only one that we have. Otherwise we give the impression that SIRF itself uses OPENMP.

Then forward to STIR (relevant option is called STIR_OPENMP).

To me, the variable name `SOURCE_DOWNLOAD_CACHE` makes no sense
```cmake
set(${proj}_SOURCE_DIR "${SOURCE_DOWNLOAD_CACHE}/sources/${proj}" )
set(${proj}_BINARY_DIR "${SOURCE_DOWNLOAD_CACHE}/builds/${proj}/build" )
set(${proj}_DOWNLOAD_DIR "${SOURCE_DOWNLOAD_CACHE}/downloads/${proj}" )
```
I'd say
```cmake
set( DOWNLOAD_CACHE ${CMAKE_CURRENT_BINARY_DIR} CACHE PATH
    "The path for downloading external source directories" )
set(SOURCE_ROOT_DIR "${DOWNLOAD_CACHE}/sources/"  CACHE PATH "blabla")
set(BUILD_ROOT_DIR "${DOWNLOAD_CACHE}/builds/"  CACHE PATH "blabla")
```
and then use those. I guess I'd make them Advanced variables.

can do this later of course.

paskino:
Why would you call it `DOWNLOAD_CACHE` when it is really the build directory?

KrisThielemans:
well ok, call it something else then that serves the function, but not `SOURCE_DOWNLOAD_CACHE` :-) In fact, if you follow my suggestion to have lots of variables, the SOURCE_DOWNLOAD_CACHE variable will nowhere be used anymore, so problem solved :-; Arguably, it serves very little purpose, as people have the option to create a build-dir already (i.e. it serves exactly the same purpose as `CMAKE_CURRENT_BINARY_DIR`)

it would still be useful to have the variable for the user to set, but it's going to create trouble with expectations of changing it but not updating all other cached variables. so either have the main one cached, or the derived ones, but not both.

KrisThielemans:
all builds are failing by the way

paskino:
it's a travis problem. The build runs fine and the test are 100% positive on travis. 
What's wrong is that travis is 
`cat SIRF-prefix/src/SIRF-build/Testing/Temporary/LastTest.log` which now doesn't exist.

https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/jobs/340409909#L6059

KrisThielemans:
:-;

keeps you busy

paskino:
To be honest, I only see a reason to set the source directory different from the `CMAKE_CURRENT_BINARY_DIR`, as users might want to fix something on the source and being able to commit back to the original packages.

The sources are currently inside the build directory, which is unnatural. So, I'd propose another thing:

```
set( SUPERBUILD_WORK_DIR ${CMAKE_CURRENT_BINARY_DIR} CACHE PATH
    "The path for binary and work directories" )
set(SOURCE_ROOT_DIR "${CMAKE_SOURCE_DIR }/../sources/"  CACHE PATH "The path for external source directories")
```

KrisThielemans:
I'm very reluctant to let it write by default outside of the build-dir. It's entirely non-standard. You don't know if they have `CMAKE_SOURCE_DIR` on a network share that could be read-only for instance. I'm alright with have just 2 variables cached of course. so let's use
```cmake
set( SOURCE_ROOT_DIR "${CMAKE_CURRENT_BINARY_DIR/sources/"  CACHE PATH
    "The path for external source directories")
```

CHANGES.md
```diff
@@ -1,3 +1,7 @@
+## New in Release 1.0
```
KrisThielemans: add a line about the current change...

SuperBuild/External_Armadillo.cmake
```diff
@@ -38,6 +40,12 @@ ExternalProject_Include_Dependencies(${proj} DEPENDS_VAR ${proj}_DEPENDENCIES)
 # Set external name (same as internal for now)
 set(externalProjName ${proj})
 
+set(${proj}_SOURCE_DIR "${SOURCE_DOWNLOAD_CACHE}/sources/${proj}" )
```
KrisThielemans: doesn't make sense as variable removed

SuperBuild.cmake
```diff
@@ -153,7 +155,7 @@ message(STATUS "PYTHON_INCLUDE_DIRS=${PYTHON_INCLUDE_DIRS}")
 #set(proj ${PRIMARY_PROJECT_NAME})
 
 # Make environment files
-set(SIRF_SRC_PATH ${SOURCE_DOWNLOAD_CACHE}/SIRF)
+set(SIRF_SRC_PATH ${SOURCE_DOWNLOAD_CACHE}/sources/SIRF)
```
KrisThielemans: should be set to `${SIRF_SOURCE_DIR}`

SuperBuild/External_Armadillo.cmake
paskino: I know. This new renaming will require changes in all external project files. I haven't finished with it yet.

paskino:
closes #16 
closes #83

PR #88, opened 2018-02-13
self host hdf5

paskino:


KrisThielemans:
all sourceforge downloads are failing now as it is down!
oh well.

PR #89, opened 2018-02-14
Add option to build UCL/petmr-rd-tools

bathomas:
Will build [UCL/petmr-rd-tools](https://github.com/UCL/petmr-rd-tools) if requested. Not that both `ITK` and [glog](https://github.com/google/glog) are required to successfully build. I have left them out of the dependency list in `External_petmr_rd_tools.cmake`. Once happy, we should add this to the VM.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -137,6 +137,10 @@ if (BUILD_siemens_to_ismrmrd)
   list(APPEND ${PRIMARY_PROJECT_NAME}_DEPENDENCIES siemens_to_ismrmrd)
 endif()
 
+if (BUILD_petmr_rd_tools)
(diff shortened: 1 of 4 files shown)


KrisThielemans:
why not have ITK as project dependency? We can build it... would have to enable the BUILD_ITK option then if this one is build.

bathomas:
Can do. Will need to enable `USE_SYSTEM_ITK` option.

KrisThielemans:
why? Are you saying `petmr-rd-tools` fails if we're building ITK itself? In that case, there's no point in enabling the dependency of course.

KrisThielemans:
The relevant option to enable our own is `USE_ITK`

KrisThielemans:
thanks. we could add 1 build in Travis (adding glog to the dependencies) but let's leave this for later. Also, we'll merge this after 1.0.0-rc.1

Issue #93, opened 2018-02-17
Number of users


casperdcl:
- [x] count superbuild clones properly (see [github-stats#viewing][gh-stats])
- [x] count VM downloads (via drupal)
- [x] count docker pulls (![](https://img.shields.io/docker/pulls/ccppetmr/sirf.svg))
  - [x] TODO: subtract travis' docker pulls (or just estimate this from [github-stats][gh-stats]'s decrements)

[gh-stats]: https://github.com/CCPPETMR/github-stats#viewing

## Ambitious

- [x] write a script (probably part of [github-stats][gh-stats]) to collate the following:
  + [x] https://github.com/CCPPETMR/github-stats/blob/CCPPETMR/SIRF-SuperBuild/CCPPETMR_SIRF_SuperBuild.json
  + [x] https://hub.docker.com/v2/repositories/ccppetmr/sirf/
  + [ ] drupal download counts? Could you have a JSON API for this @paskino 
  + [x] travis' docker pulls?

## original description

We'd like to estimate the number of users. Sum number of:

+ (SuperBuild clones) - (travis non-docker jobs) - 2*(travis docker jobs) - (a little extra for devs locally cloning?)
+ (docker pulls) - (travis docker jobs)
+ VM binary downloads

The main problem I see is (SuperBuild clones) is only accessible at https://github.com/CCPPETMR/SIRF-SuperBuild/graphs/traffic if you are logged in to github (not publicly accessible via a JSON API) so I can't get that data to generate a badge.

So... would it be OK to generate that data ourselves using `cmake`? By that I mean something like

```cmake
file(DOWNLOAD "http://caspersci.uk.to/cgi-bin/hits.cgi?q=sirf-superbuild"
  "${CMAKE_BINARY_DIR}/clones.svg")
# file(REMOVE "${CMAKE_BINARY_DIR}/clones.svg")
```

to increment our counter, and then get `.travis.yml` to call

```bash
wget https://caspersci.uk.to/cgi-bin/hits.cgi?q=sirf-superbuild-travis -o /dev/null -b
```

a few times to decrement... that way I can give you something like: ![](https://caspersci.uk.to/cgi-bin/hits.cgi?q=sirf-superbuild&m=sirf-superbuild-travis&label=installs&a=display&l=https://avatars2.githubusercontent.com/u/16674841?s=32&amp;v=4)

casperdcl:
@KrisThielemans ?

[...]

PR #98, opened 2018-03-02
Usercount

casperdcl:
Related to #93

- [x] host our own database of clones (https://github.com/CCPPETMR/github-stats)
- [x] get travis to automatically update this using github's API (which only has the last fortnight's worth of clones)
- [x] automatically exclude counting travis clones

Will need to add a 3 lines to SIRF/.travis.yml to do the same there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -172,6 +172,22 @@ before_install:
  # ensure python bin dir exists (and coverage dependencies installed)
  - $PY_EXE -m pip install --user -U nose codecov coveralls
  - |
+    # for counting clones, excluding ours
+    $PY_EXE -m pip install --user -U argopt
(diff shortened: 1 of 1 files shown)


KrisThielemans:
looks good to me, but can you clarify where would the GITHUB_API_TOKEN come from, and who would need to create it (me?)

casperdcl:
already done, it's a secure env var in travis. The token itself belongs to @ccp-petmr-codebot and allows full access to its repos.

KrisThielemans:
ok. interesting how it picks that up (or possibly Travis is run via that user, I cannot remember).

Looks fine to me otherwise, so feel free to merge. Thanks!

casperdcl:
No, I set this all up manually, but it should be future-proof and fully controlled by @ccp-petmr-codebot rather than my account.

Issue #97, opened 2018-03-02
travis osx python3


casperdcl:
build failing due to brew, probably easy to fix:

https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/jobs/344891590

rijobro:
I think that for OSX with python3, this could be solved by replacing `brew install python3` with `brew upgrade python`

KrisThielemans:
correct. there's probably more to it. However, I've recently fixed this in STIR. same should work here.
https://github.com/UCL/STIR/commit/616cb0f382fd8668dbc51f59a1603447b58d0582

PR #100, opened 2018-03-07
Python#97

bathomas:
Working on #97.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,7 +130,7 @@ before_install:
         brew upgrade python || travis_terminate 1
         brew install python3 || travis_terminate 1
 
-        TARGETPYVER=`find /usr/local/Cellar/python3/* -type d  -maxdepth 0 | sed 's#.*/##' | sort -V | tail -n 1`
+        TARGETPYVER=`find /usr/local/Cellar/python3/ -type d  -maxdepth 1 | sed 's#.*/##' | sort -V | tail -n 1`
         if [ -z $TARGETPYVER ]; then
           travis_terminate 1
         fi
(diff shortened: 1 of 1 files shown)


KrisThielemans:
now fails with
```
/Users/travis/.travis/job_stages: line 57: /usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6/bin/python3.6: No such file or directory
```
clearly, the following in .travis.yml are unsafe
```
        export PYINST=/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions
        export PY_EXE=$PYINST/$PYMVER.6/bin/python3.6
```

rijobro:
from the log:
`==> Upgrading 1 outdated package, with result: python 3.6.4_3`
So its looking for 3.6.4_2 but should be looking for 3.6.4_3. So yes, it looks dangerous to hardwire the python path.

.travis.yml
```diff
@@ -127,8 +127,15 @@ before_install:
         export BUILD_FLAGS="$BUILD_FLAGS -DPYTHON_LIBRARY=$PYINST/2.7/lib/libpython2.7.dylib"
         export BUILD_FLAGS="$BUILD_FLAGS -DPYTHON_INCLUDE_DIR=$PYINST/2.7/include/python2.7"
       else
+        brew upgrade python || travis_terminate 1
         brew install python3 || travis_terminate 1
-        export PYINST=/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions
+
+        TARGETPYVER=`find /usr/local/Cellar/python3/* -type d  -maxdepth 0 | sed 's#.*/##' | sort -V | tail -n 1`
```
KrisThielemans: ok. better. but you cannot have the *. find needs a single argument for the directory. just delete it.

KrisThielemans:
I've fixed this  in #118 in a related manner, but simpler for the libraries. I didn't use the `find` trick which might come in handy later.

Issue #63, opened 2017-12-20
Building with rpaths on OSX High Sierra


rijobro:
With OSX, dynamic libraries can be referred to as (using the example of libboostfilesystem.dylib):
1. `libboostfilesystem.dylib`,
2. `@rpath/libboostfilesystem.dylib`, or
3. `/absolute/path/to/libboostfilesystem.dylib`.

Compilation against the first of these three causes linking errors on my version of OSX. This is possibly due to changes to SIP for High Sierra (or possible a problem specific to my machine). 

Possible solutions include: 
1. using install_name_tool to change the first option to the second or third, or
2. linking against static libraries.

Hopefully this issue will sort itself out over time.

KrisThielemans:
@rijobro, I think this can be closed?

rijobro:
The option, `SHARED_LIBS_ABS_PATH`, means that all dynamic libraries compiled by CMake will have the full paths, and so the problem is resolved. However, for dependencies that are not compiled with CMake (e.g., Boost, which is compiled with boostrap), the problem remains.

A short-term fix is to use `brew install boost` (or `brew upgrade boost`, if they're not using the current version, 1.66).

As for long-term, I'm not so sure. Boost isn't compiled with CMake; what else isn't? We would have to find fixes for all of those...

KrisThielemans:
I was wondering if this wouldn't be fixed if we'd let the SuperBuild use boost 1.66. I saw no reason not to, until I found that [boost 1.66 doesn't compile on Travis OSX](https://github.com/boostorg/build/issues/279), and therefore presumably some versions of OSX. Sigh.

I therefore suggest that on OSX, we currently set the USE_SYSTEM_BOOST to ON by default, and document the `brew` stuff in our OSX installation instructions.

PR #104, opened 2018-03-26
use system boost by default because of #63

paskino:


rijobro:
@KrisThielemans - as requested, the info telling Mac users to upgrade to Boost 1.66 is given [here](https://github.com/CCPPETMR/SIRF/wiki/SIRF-SuperBuild-on-MacOS#2-boost-). I wonder if it's not in an obvious enough location?

KrisThielemans:
sorry to be late. I thought we'd only do this for OSX and Windows, but keep it OFF elsewhere. @bathomas, @casperdcl any opinions?

paskino:
No worries. I misunderstood your email. 
I'll add a switch `if not UNIX`

KrisThielemans:
@rijobro we'll now require it, so it needs to be moved up to "requirements". sorry

bathomas:
Seems unnecessary to default to `ON` for an OS that doesn't have a problem. SIRF(-SuperBuild) is straightforward to use on Ubuntu. Also, the packaged version would be older (1.54?) on Travis.

paskino:
The Superbuild is actually building Boost 1.63. Wouldn't it be sufficient to upgrade the built (required) version to 1.66?

rijobro:
It wouldn't be sufficient - Brew's version of 1.66 solves the problem on OSX (thus why `USE_SYSTEM_Boost` works). 

But the problem is still there with the boostrapped version 1.66 of boost (the one that will be installed by the superbuild).

bathomas:
As @KrisThielemans said, I think `USE_SYSTEM_Boost` should be `OFF` for Linux systems and `ON` for the others.

paskino:
OK, then I'll add the switch for non Unix. 
Can we fix our own compilation of Boost on MacOSX?

rijobro:
I looked into this before. We install Boost with `./bootsrap.sh` and then `./b2 install`, so in one of those two we would have to add an option that would install the absolute paths to the dynamic libraries. So far, I haven't found a way of doing that.

rijobro:
Or maybe, add the switch `if OSX` in CMake, and build Boost with static libs if true?

paskino:
To be clear, on OSX the only way to have the SuperBuild working is to install Boost 1.66 with `brew`

paskino:
@rijobro what about this https://stackoverflow.com/questions/3431619/how-to-force-boost-to-use-rpath and especially `hardcode-dll-paths=true dll-path=/custom/path/lib`

rijobro:
I tried that before with no luck (might be out of date). I'll try again and let you know.

rijobro:
Tried again with no effect (maybe outdated?)

PR #105, opened 2018-03-28
Update stir version

paskino:


casperdcl:
I guess there's no docker issue now? The devel build is still failing but that's because of gadgetron

KrisThielemans:
yes. docker build is fine now (I haven't tried it of course).

PR #106, opened 2018-04-11
Fix python dest

paskino:
The `PYTHON_DEST` variable was always overwritten to `${SIRF_Install_Dir}/python'

This PR goes along with https://github.com/CCPPETMR/SIRF/pull/160

Issue #102, opened 2018-03-11
install demos and example data


KrisThielemans:
at the moment, the demos and data are in the SIRF clone. This is non-standard, as you'd expect that an install has everything.

This would be easy enough to do. For demos themselves hardly take space, data is currently ~100MB.

KrisThielemans:
should be in `$INSTALL/share/SIRF` I believe. This is a SIRF issue really. @paskino, please create an issue there. We could make it optional if people think this is necessary. `env_ccppetmr.sh` should know about this then.

KrisThielemans:
migrated to https://github.com/CCPPETMR/SIRF/issues/181

PR #112, opened 2018-04-26
Conda

paskino:
initial working recipe for linux.

OSX would need tweak on build.sh. 

The branch should be aligned with setup_py #109


casperdcl:
1. merge https://github.com/CCPPETMR/SIRF/pull/160
2. merge #109
3. rebase this and merge

close #106?

paskino:
I agree with you.

KrisThielemans:
should we add a conda build to Travis? Maybe it's already done via the docker build.

that'd allow you to test on OSX.

KrisThielemans:
Data and examples is #102.

evgueni-ovtchinnikov:
@casperdcl: so far as i can see examples in MR/interactive require simulated_MR_2D_cartesian.h5, which is not in https://github.com/CCPPETMR/SIRF_data/tree/master/examples

@ckolbPTB: why did you not use existing_filepath function? the error message would be more meaningful

also, the graphics in coil_combination.py crashes python when run from command line (spyder survives)

conda-recipe/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - vc 14 # [win and py36] 
+    - vc 9  # [win and py27]
+    - numpy
+    - scipy
+    - matplotlib
+    - docopt
+    - h5py
+    - libxml2
+    - psutil
+    - nose
```
KrisThielemans: `nose` and `coverage` not needed for conda package? (they probably don't make sense)

conda-recipe/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - python 3.6 # [py36]
+    - python 2.7 # [py27]
+    - cmake >=3.1
+    - vc 14 # [win and py36] 
+    - vc 9  # [win and py27]
+    - numpy
+    - scipy
+    - matplotlib
+    - docopt
+    - h5py
```
KrisThielemans: 99% sure we don't need this

conda-recipe/build.sh
```diff
@@ -0,0 +1,69 @@
...
+#    exit 1
+#fi  
+
+mkdir "$SRC_DIR/build"
+mkdir "$SRC_DIR/SIRF-SuperBuild"
+cp -rv "$RECIPE_DIR/../" "$SRC_DIR/SIRF-SuperBuild"
+
+cd $SRC_DIR/build
+
+echo "$SRC_DIR/ccpi/Python"
```
KrisThielemans: really?

conda-recipe/build.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+#
+# Script to install/update the CCP-PETMR VM. It could also be used for any other system
+# but will currently change your .sirfrc. This is to be avoided later on.
```
KrisThielemans: incorrect comment

conda-recipe/build.sh
```diff
@@ -0,0 +1,69 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+#=========================================================================
+
+#if [ -z "$SIRF_VERSION" ]; then
+#    echo "Need to set SIRF_VERSION"
+#    exit 1
+#fi
```
KrisThielemans: until we fix this, will need to add it to the checklist

conda-recipe/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - psutil
+    - fftw
+    - libgcc-ng
+    - libstdcxx-ng
+    - openblas
```
KrisThielemans: not for build?

KrisThielemans (changes requested):
could probably easily split ismrmrd and gadgetron off...

conda-recipe/build.sh
paskino: Really what?

conda-recipe/meta.yaml
paskino: They are needed during the build

conda-recipe/build.sh
KrisThielemans: ccpi...

conda-recipe/meta.yaml
KrisThielemans: not convinced. they are needed for the testing. but are you testing here?
not a problem to leave them in if it helps of course.

paskino:
I guess we'll end up creating several recipes which create the packages sirf depends on, and sirf.
As we discussed, we could pass a parameter to the superbuild and build only part of it. Now I don't remember which parameter is exactly.

conda-recipe/meta.yaml
paskino: It wouldn't build otherwise

conda-recipe/build.sh
paskino: You know, I'm torn

KrisThielemans:
currently you'd run CMake as usual. then just say `make gadgetron` or whatever. A bit wasteful, but minimal effort...

KrisThielemans:
this is fine-tuning of course. leave it for later!

conda-recipe/meta.yaml
KrisThielemans: well... in that case, we need it! :-)

conda-recipe/meta.yaml
casperdcl: I feel like there's a difference between build, install (runtime), and test. I'd think nose and coverage fall into the last category, and things like h5py for install? Is there a way to specify these three categories for conda? I'm sure at least two are possible. At least split into sections with comments

paskino: Indeed there is. Currently The tricky bit is afaik that the install phase of the external packages is run during the make phase.

paskino:
I wouldn't mind issuing cake 3 or 4 times!

casperdcl:
> I guess we'll end up creating several recipes which create the packages sirf depends on, and sirf.

helps that STIR is now in `conda-forge`

KrisThielemans:
> helps that STIR is now in conda-forge

First need to resolve https://github.com/UCL/STIR/issues/211

paskino:
Directories to pass to `External_SIRF.cmake` are defined 

* For STIR [here](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/master/SuperBuild/External_STIR.cmake#L128L130)
* NIFTYREG [here](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/master/SuperBuild/External_NIFTYREG.cmake#L72L74)
* ITK (needed by [STIR](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/master/SuperBuild/External_STIR.cmake#L109)) where?

conda/boost/build.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
KrisThielemans: I would try to avoid building boost at all costs. Use the conda dependency

conda/hdf5/build.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
KrisThielemans: let's not build hdf5

conda/ismrmrd/build.sh
```diff
@@ -0,0 +1,72 @@
...
+cd $SRC_DIR/build
+
+
+#site-packages ${SP_DIR}/sirf
+cmake ../SIRF-SuperBuild \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DPYTHON_DEST_DIR=${PREFIX}/python\
+    -USIRF_URL \
+    -USIRF_TAG \
+    -DSIRF_TAG=v2.1.0\
```
KrisThielemans: let's not put explicit versions in any of the conda build.sh. In any case, SIRF version is irrelevant in ismrmrd/build.sh

conda/sirf/build.sh
```diff
@@ -0,0 +1,82 @@
...
+    -DBUILD_siemens_to_ismrmrd=Off \
+    -DUSE_SYSTEM_GTest=On\
+    -DUSE_ITK=ON\
+    -DUSE_SYSTEM_ITK=OFF\
+    -DUSE_SYSTEM_NIFTYREG=OFF\
+    -DCONDA_BUILD=On
+
+make -j2 ITK
+make -j2 NIFTYREG
+make -j2 STIR
```
KrisThielemans: no need for these.

conda/stir/build.sh
```diff
@@ -0,0 +1,76 @@
...
+    -DUSE_SYSTEM_Armadillo=On \
+    -DUSE_SYSTEM_ISMRMRD=ON\
+    -DUSE_SYSTEM_STIR=Off\
+    -DUSE_SYSTEM_FFTW3=On \
+    -DUSE_SYSTEM_HDF5=ON \
+    -DBUILD_siemens_to_ismrmrd=Off \
+    -DUSE_SYSTEM_GTest=On\
+    -DCONDA_BUILD=On
+
+make  -j1 Gadgetron
```
KrisThielemans: in stir/build.sh?

KrisThielemans (changes requested):
I haven't checked this really, but I'm confused about the various `build.sh` files, and then `sirf/build.sh` building other stuff anyway.

KrisThielemans:
STIR 4.0.1 conda might work ok. There's https://anaconda.org/inati/ismrmrd but it's out-of-date. Might be worth checking with them.

KrisThielemans:
By the way, I'm not 100% sure what the strategy here was/is, but I think we should have a SIRF-only conda package (with its dependencies, most obtained via conda). I would at least initially not attempt gadgetron-via-conda.

KrisThielemans:
See https://github.com/gadgetron/gadgetron/pull/1036 for Gadgetron's `conda` efforts

conda-recipe/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - libxslt
+    - armadillo
+    - gtest
+
+
+  run:
+    - vc 14 # [win and py35]
+    - vc 9  # [win and py27]
+    - python 3.6 # [py36]
+    - python 2.7 # [py27]
```
KrisThielemans, line 42: drop this of course

paskino:
hugely obsolete work

Issue #134, opened 2018-08-08
Pin DEVEL_BUILD to recent gadgetron


KrisThielemans:
Gadgetron master is tripping us up rather regularly unfortunately. I feel we should fix `DEVEL_BUILD=On` to use a specific (but recent) gadgetron version as opposed to its master. We can then update the `GADGETRON_TAG` occasionally via a SIRF-SuperBuild PR. Any objections?

we could do the same for STIR if it becomes necessary.

KrisThielemans:
We're now getting failures on `DEVEL_BUILKD`
```
CXX COMPILER VERSION: GNU : 4.8.4
CMake Error at CMakeLists.txt:43 (message):
  Gadgetron requires GCC version >= 6.0
```
that'll be fun on travis etc.

In any case, this reinforces the motivation for doing the above. @DANAJK @ckolbPTB @johannesmayer  any objections? Please suggest a commit if you can

DANAJK:
I have no objections. Unfortunately I cannot suggest a specific commit for Gadgetron.

KrisThielemans:
see #146

PR #145, opened 2018-08-24
Update ACE package

paskino:
Reflects updates done on ACE.
Uses Gadgetron version that compiles with GPU support.

### DIFF
--- a/SuperBuild/External_ACE.cmake
+++ b/SuperBuild/External_ACE.cmake
@@ -60,14 +60,17 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
     DOWNLOAD_DIR ${${proj}_DOWNLOAD_DIR}
     STAMP_DIR ${${proj}_STAMP_DIR}
     TMP_DIR ${${proj}_TMP_DIR}
-    INSTALL_DIR ${libACE_Install_Dir}
 	
-    INSTALL_COMMAND make ACE 
     CMAKE_ARGS
-        -DCMAKE_INSTALL_PREFIX=${libACE_Install_Dir}
+        -DCMAKE_INSTALL_PREFIX=${Gadgetron_Install_Dir}
+	-DLIBRARY_DIR=${libACE_Install_Dir}/lib
+	-DINCLUDE_DIR=${libACE_Install_Dir}/include
+
+    INSTALL_COMMAND make ACE
     DEPENDS
         ${${proj}_DEPENDENCIES}
   )
+
     set(${proj}_ROOT        ${${proj}_SOURCE_DIR})
     set(${proj}_INCLUDE_DIR ${${proj}_SOURCE_DIR})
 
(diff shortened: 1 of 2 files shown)


KrisThielemans:
I'm somewhat reluctant to point to our own gadgetron fork, as it gives the impression we're deviating from them. I vote for pointing to the main gadgetron repo somewhere after acceptance of your PR. @DANAJK , @ckolbPTB @johannesmayer any suggestions/opinions?

DANAJK:
Similar to #134 , I have no objection to pointing to a Gadgetron commit that works for us. Hopefully more recent that one year though.

paskino:
I tried to build Gadgetron after the PR was merged and it fails. 
I had a first error coming from the CXX flag `-Werror=terminate` which isn't recognised. 

Fixed that (requires changing Gadgetron CMakeLists.txt) and I got to 
```
/home/sirfuser/devel/buildgadgetron/sources/Gadgetron/toolboxes/mri_core/mri_core_data.h:5:30: fatal error: ismrmrd/waveform.h: No such file or directory
compilation terminated.
```

DANAJK:
waveforms are something that is being added to ISMRMRD (to support ECG information as an example).
Maybe you need to update ISMRMRD too?

KrisThielemans:
we should follow the same strategy for ISMRMRD as in #134  I think: occasional updates, with testing first. if we need ECG (or any other schanges in ISMRMRD) now or not I have no idea. As long as it doesn't break things, it's fine for me of course.

KrisThielemans:
ah sorry. I hadn't seen Edo's error. Seems Gadgetron doesn't check the ISMRMRD version properly then...

Then it needs updating ISMRMRD of course.

paskino:
I'll see if I find a couple of Gadgetron/ISMRMRD which work for us. 
The `-Werror=terminate` problem might mean that there is a minimum requirement on gcc version, and the one we have isn't sufficient.

KrisThielemans:
I guess do a `git blame` to find out when they introduced that option. Not nice if it isn't guarged by appropriate checks on compiler.

paskino:
OK, the new Gadgetron/master requires at least GCC version 6. 

Ubuntu 16.04 has 5.4. On the VM it could be possible to install gcc version 6 following [this link](https://askubuntu.com/questions/746369/how-can-i-install-and-use-gcc-6-on-xenial/746480)

I see a few options:

1. Build a new VM with Ubuntu 18.04 which ships [GCC 7.3.0](https://packages.ubuntu.com/bionic/gcc) 
2. Add the new unofficial repository to get GCC 6 onto the current VM (Ubuntu 16.04)
3. Use conda
4. Keep outdated Gadgetron in the short term.

Although option 3 is still in a [work-in-progress](https://github.com/CCPPETMR/SIRF-SuperBuild/pull/112) phase I strongly believe that we should pursue that direction.

paskino:
They introduced a check of the availability of `-Werror=terminate` (which I incidentally couldn't find documented anywhere) in the latest Gadgetron master branch. But along with that, they check minimum version of GCC >= 6.

KrisThielemans:
This discussion isn't really about this PR. I've created a new Issue for it: #146 

This PR does 2 things: change ACE stuff, and change Gadgetron Tag. Those are independent, and the Gadgetron stuff is preventing us merging the ACE changes.

@paskino, please revert Gadgetron changes, change the title of the PR, and then squash-merge.

Issue #147, opened 2018-10-07
FFTW3 and ACE location can only be set via environment variable


KrisThielemans:
FindFFTW3.cmake (supplied in Gadgetron, ISMRMRD and SIRF, which are all identical) use `$ENV{FFTW_ROOT_DIR}`. Therefore, setting the `FFTW_ROOT_DIR` CMake variable has no effect. Instead, when using `USE_SYSTEM_FFTW3=OFF`, the user has to set the environment variable before building the superbuild (note that setting it when running CMake the first time will not have any effect).

The same holds for `ACE_ROOT`.

@evgueni-ovtchinnikov , please add this to our installation documentation (wike etc) and close this issue when done.

Notes
------

- As CMake doesn't come with FFTW3 support, it is currently necessary to have FindFFTW3.cmake reproduced 3 times. sigh.

- This is poorely documented in [FindFFTW3.cmake](https://github.com/gadgetron/gadgetron/blob/6abefd6d4576641c44b2ac5db3d4b3b2ecce90eb/cmake/FindFFTW3.cmake#L5), which could be fixed by a PR to Gadgetron, ISMRMRD and SIRF. Similarly, we could fix `FindFFTW3.cmake` to also use the CMake variable.

- I have tried to use ` set(ENV{FFTW3_ROOT_DIR} bloe)`
in our `External_FindFFTW.cmake`. This however doesn't pass the environment variable to the CMake instances for Gadgetron etc.

- By the way, when using `USE_SYSTEM_FFTW3=OFF`, CMake currently does find our own installation even if the  `FFTW3_ROOT_DIR` env variable (as `find_library` etc give precedence to `MAKE_PREFIX_PATH` over `HINTS` ).

KrisThielemans:
@evgueni-ovtchinnikov can you then add something like this to the doc as well? Feel free to improve:
> CMake does come with FindArmadillo.cmake but it currently (at least up to CMake 3.12) has no variable to specify its location at all. This implies that when using `USE_SYSTEM_ARMADILLO=On`, you have to install armadillo installed in a system location, unless some extra work is done. See [this post on stackoverflow](https://stackoverflow.com/questions/35304513/cmake-find-armadillo-library-installed-in-a-custom-location) for some suggestions, which we haven't tried.

KrisThielemans:
[...]

PR #148, opened 2018-10-07
Mark as superbuild and other fixes for badly supported packages

### LINKED ISSUE #70: usage of `mark_as_superbuild`
I don't think we use this properly.

# `SuperBuild.cmake` should add some variables to all subprojects

Indeed, it tries to do this (could do it for more of them I think). However, [it says](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/750467ab5c3024003923fea48e8a42dd83574c10/SuperBuild.cmake#L53)
```cmake
mark_as_superbuild(
   PROJECTS ALL_PROJECTS
   VARS CMAKE_GENERATOR:STRING CMAKE_GENERATOR_PLATFORM:STRING CMAKE_GENERATOR_TOOLSET:STRING
        CMAKE_C_COMPILER:FILEPATH CMAKE_CXX_COMPILER:FILEPATH
        CMAKE_INSTALL_PREFIX:PATH
)
```
I think this has to be 
```cmake
mark_as_superbuild(
   ALL_PROJECTS
 ..
)
```
No idea why it would work now... but I think it does as [I put it in](733a35985d7ffbee9ad9ecdf0e79262be6292269) :-) 

# All the external projects use it to pass CMake variables on
code is
```cmake
mark_as_superbuild(
    VARS
        ${externalProjName}_DIR:PATH
    LABELS
      "FIND_PACKAGE"
  )
```
This has 2 problems as far as I can see:
- It is only correct if the project indeeds sets/need the `*_DIR` variable (many do not, e.g. `GTEST_ROOT`).
- According to the doc, `mark_as_superbuild` adds this variable then only to the top-level project, which in our case is `CCPPETMR`, but not to the projects where they are actually needed.

The 2nd item seems to imply these lines are currently effectively ignored. Hence the need for us to specify all the variables by hand again.

I wonder if we should use `ALL_PROJECTS` here as well, although then we're going to add variables where we actually don't need/use them. A better way would be to know to which projects we need to add them, but then we introduce dependencies on "downstream" projects, which seems a bad idea.

Probably a better solution is to forget about `mark_as_superbuild` in the sub-projects, and to introduce a variable with all CMake args that need to be set if someone wants to use this project (but that seems exactly what `mark_as_superbuild` seems to aim to provide).

# Summary
I'm confused...

Here's [an old-ish post with some info](https://cmake.org/pipermail/cmake/2014-June/057738.html). It refers to old Slicer stuff, but [here's a recent link](https://github.com/Slicer/Slicer/blob/master/SuperBuild/External_ITKv4.cmake). Note that this also has some interesting stuff with `unset` and `ExternalProject_SetIfNotDefined` which might be relevant to #51.

KrisThielemans:
This PR cleans up a few files which used `mark_as_superbuild` or other mechanisms to pass locations to other packages. For older packages this failed because their `Find*.cmake` don't do what we thought they did.

See #147. This also partially resolves #70.


KrisThielemans:
@paskino, this seems ready to go. One failure on Travis due to Gadgetron and gcc 6. 
Could you have a look and add your review?

paskino:
The build that fails on [travis](https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/builds/438522270?utm_source=github_status&utm_medium=notification) is using `DEVEL_BUILD=ON` therefore crashing on Gadgetron/master which requires GCC >6.

I tested and confirm it works. Merging.

Issue #14, opened 2017-04-29
Using different compilers does not always go to dependencies


KrisThielemans:
In bash, you can normally say

    CC=gcc-5 CXX=g++-5 cmake .

to use a non-default compiler. However, these settings of CC do not get passed on to (some?) dependencies with the superbuild. It currently only works if you do

    export CC=gcc-5
    export CXX=g++-5
  cmake .

KrisThielemans:
This might help for the CMake projects (but not for configure etc)

http://stackoverflow.com/questions/32744788/forwarding-current-compiler-to-externalproject?rq=1

KrisThielemans:
This should no longer be a problem for CMake projects by using `mark_as_superbuild`. So I guess only configure projects affected. The suggested solution should work, so we'll postpone.

KrisThielemans:
The suggested solution of setting `CC` and `CXX` does not work for boost as Boost.Build ignores this and  uses the system gcc anyway. It should be solvable by creating a `user-config.jam`, see [an answer by Dave Riedel here](https://lists.boost.org/boost-users/2013/01/77132.php)
```
I have several different versions of gcc I need to work with so,
I have done this for a long time by modifying my user-config.jam as follows:
using gcc : : /<path to custom gcc>/bin/g++ : <linkflags>"-Wl,-rpath
-Wl,/<path to custom gcc>/lib64" ;

You need the rpath so that any shared libraries will find the correct
gcc runtime libraries. 
```
This would conceivably be possibly to do [via CMake here](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/b16e1499107de97650d9c3c90cc3603dd54d0306/SuperBuild/External_Boost_configureboost.cmake#L26), but it isn't trivial to get the complete and correct paths.  The `linkflags`  are presumably gcc specific as well.

we might still have to run `./b2 --toolset=gcc`.

Boost `bootstrap.sh` seems to have a special toolset `cc` that might use `CC` environment variable, but @evgueni-ovtchinnikov  and I couldn't get that to work.

KrisThielemans:
Specifying compilers is actually not enough. We need to port C++ versions as well. See https://github.com/CCPPETMR/SIRF-SuperBuild/issues/161

PR #157, opened 2018-12-04
[TRAVIS] fix DEVEL_BUILD

KrisThielemans:
required for recent Gadgetron


KrisThielemans:
@casperdcl how do we switch the docker devel build to use gcc6?

KrisThielemans:
something wrong in our Travis file as it ignores the `CC=gcc-6` settings. Checking [a log](https://api.travis-ci.org/v3/job/463585998/log.txt) I find
```
$ export EXTRA_BUILD_FLAGS="-DDEVEL_BUILD=ON -DUSE_SYSTEM_Boost=OFF -DUSE_SYSTEM_FFTW3=OFF -DUSE_SYSTEM_HDF5=OFF -DUSE_SYSTEM_SWIG=OFF"
$ export CC=gcc-6
$ export CXX=g++-6
$ export PYMVER=2

$ export TRAVIS_COMPILER=g++
$ export CXX=g++
$ export CXX_FOR_BUILD=g++
$ export CC=gcc
$ export CC_FOR_BUILD=gcc
```
Anyone? @casperdcl ?

paskino:
I used to set `CMAKE_CXX_COMPILER` and `CMAKE_C_COMPILER` variables issuing CMake on our cluster to get it pick the right compiler.

paskino:
First, I don't find where this variables are set. 

However in the [`.travis.yml`](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/travis_gcc6/.travis.yml#L41) I see
```
addons:
      apt:
        sources:
         - ubuntu-toolchain-r-test
        packages:
            - g++-6
```
which seems to imply that we install the `g++-6` package from `ubuntu-toolchain-r-test`.  I do not find that package at [their website](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test/+packages), only `gcc-6`. Maybe I'm totally wrong. 

At any rate the `gcc-6` package does not contain a binary `g++-6`.

paskino:
On my 16.04 VM
```
sudo add-apt-repository ppa:ubuntu-toolchain-r/test
sudo apt-get update
sudo apt-cache search g++-6

> clang-6.0 - C, C++ and Objective-C compiler
> clang-6.0-doc - C, C++ and Objective-C compiler - Documentation
> clang-6.0-examples - Clang examples
> libclang-6.0-dev - clang library - Development package
> python-clang-6.0 - Clang Python Bindings
```

KrisThielemans:
I have no idea about the toolchain-r, but g++-6 is installed ok. we can even get g++-8 there. you can see it in the log (link above).

The problem is just the setting of the environment variables in Travis. we set them for each configuration (e.g. [here](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/894010dc8d67d0e6c87aee5b424162363eee4823/.travis.yml#L71), as the compilers are different for each config. That's ok as you can see in the log and my quote in the first post. However, then Travis overrides them with something else, presumably because of how our yml file is. (We needed to [set the config to python](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/894010dc8d67d0e6c87aee5b424162363eee4823/.travis.yml#L66) to get that side to work).

KrisThielemans:
@paskino, please revert this /82b09356a52acafd75bb6333a141b8f9bfbe4146 commit.

NOT all configs use gcc6. we will want to be able to use different compilers.

Also, using the CMake flags doesn't transport to boost, fftw etc, see #14.

paskino:
Of course, but how would I check if the changes work?

KrisThielemans:
remaining failures are due to a problem with Gadgetron master https://github.com/gadgetron/gadgetron/pull/715

casperdcl:
just to note that only 2 docker images are built: standard and devel (ubuntu 16.04, gcc5)

see default flags here: https://github.com/CCPPETMR/SIRF-SuperBuild/blob/c3a2c065e3ff0358b8b2c3dbf9961268b2f03cc0/docker/Dockerfile#L78-L82

PR #171, opened 2018-12-09
CI statistics fixes

casperdcl:
Fixes issues with PRs failing due to not having stats access permissions


casperdcl:
Yes, should fix both. As a side note, forked PR builds still pull the source (this) repo, so will contribute to the stats. Maybe we don't mind this too much?

PR #182, opened 2019-01-13
Add CMake options for testing sub-projects

### LINKED ISSUE #6: what to do with BUILD_TESTING?
Several sub-projects have BUILD_TESTING.
We could provide several options (no testing, test only SIRF, test everything). In some sense however the "test only SIRF" is flawed as the dependencies need to work correctly. But running all the tests is going to take a looong time.

I guess we could have a CMake option for every project dependency (could maybe be done semi-automatically: add an option BUILD_TESTING_${proj}, which then sets its BUILD_TESTING).

This would allow the developer to choose what tests to run.

KrisThielemans:
Fixes #6

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -285,6 +285,3 @@ configure_file(env_ccppetmr.csh.in ${CCPPETMR_INSTALL}/bin/env_ccppetmr.csh)
 
 # add tests
 enable_testing()
-add_test(NAME SIRF_TESTS
-         COMMAND ${CMAKE_CTEST_COMMAND} -C $<CONFIGURATION>
-         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/builds/SIRF/build/)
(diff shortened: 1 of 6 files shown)


KrisThielemans:
This seems to work fine. It was a bit more complicated than hoped for as I needed to put in specific code for ISMRMRD, but otherwise ok.

One question: should we call this `BUILD_TESTING_${proj}` or `${proj}_BUILD_TESTING`? The advantage of the latter is that it'll be grouped with other variables for the project, but of course the advantage of the former is that they're grouped together. I have currently used the former as that's what I said in #6. I'm slightly leaning towards the latter though. Opinions?

KrisThielemans:
ISMRMRD tests are fast (need a bit of compilation), so I've left them `ON`. Same for Gadgetron (currently running its unit tests only. The integration tests need a lot of data and take a long take). However, for Gadgetron, we presumably could remove the GTest dependency of Gadgetron if we don't want to test it, which would save some time.

As with our current `version_config.cmake` we pin ISMRMRD and Gadgetron versions, for Travis, arguably we should test those only  occassionaly. This could be done by having the tests `OFF` in the `master` .travis.yml, but have a specific branch `RunAllTests` where they are `ON`. Whenever we do a major change, we could then merge `master` onto that.

I suggest we do that when the tests are starting to take more time.

PR #196, opened 2019-01-23
If building own version of boost, try to minimise looking for other versions

rijobro:
Should be ok, but untested


rijobro:
> Also, we should set these in External_Boost.cmake really (in the spirit of modularity)

Done. I put them in the root because I was worried about scope.

rijobro:
Merging this as Travis is green.

PR #205, opened 2019-02-12
Fix OSX Gadgetron

rijobro:
- Make ISMRMRD and Gadgetron versions match for both devel and non devel  (devel build was using older versions than non-devel, which didn't make much sense)
- Update Gadgetron version (fixes openblas problem on OSX)
- For OSX, need to pass openblas lib and include paths (for Travis, too)


SuperBuild/External_Gadgetron.cmake
```diff
@@ -64,6 +64,14 @@ if (MKL_FOUND)
   endif ()
 endif ()
 
+if (APPLE)
```
KrisThielemans: better to use CMake's [FindBlas](https://cmake.org/cmake/help/v3.14/module/FindBLAS.html), or is this not used by Gadgetron?

.travis.yml
```diff
@@ -150,6 +150,8 @@ before_install:
     if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
       brew update
       BUILD_FLAGS="$BUILD_FLAGS -DSHARED_LIBS_ABS_PATH=ON"
+      brew install openblas
+      EXTRA_BUILD_FLAGS="$EXTRA_BUILD_FLAGS -DCBLAS_INCLUDE_DIR=/usr/local/Cellar/openblas/0.3.5/include -DCBLAS_LIBRARY=/usr/local/Cellar/openblas/0.3.5/lib/libblas.dylib"
```
KrisThielemans: would have to get version number from somewhere at some point. if you cannot do this now, raise an issue for later

version_config.cmake
```diff
@@ -85,6 +85,14 @@ set(ITK_TAG v4.13.1)
 set(NiftyReg_URL https://github.com/KCL-BMEIS/niftyreg.git )
 set(NiftyReg_TAG 731a565bd42ca97ff5968adb1c06133ea72f0856 )
 
+## ISMRMRD
+set(DEFAULT_ISMRMRD_URL https://github.com/ismrmrd/ismrmrd )
+set(DEFAULT_ISMRMRD_TAG v1.4.0)
+
+## Gadgetron
+set(DEFAULT_Gadgetron_URL https://github.com/gadgetron/gadgetron )
+set(DEFAULT_Gadgetron_TAG e03b8a254eb25ff158175ac453500aa8b3c6199c)
```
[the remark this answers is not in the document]
rijobro: I think so. I'm not sure when the change was made, but at some point after 3.17.0 it became possible to enter the paths to `blas`, which is what we need to avoid using the default Apple one that seems to be causing the problems.

SuperBuild/External_Gadgetron.cmake
rijobro: ok, now using `find_package(blas)` (but still only for `APPLE`).

.travis.yml
rijobro: done.

rijobro:
Not ready for merge yet, waiting on the solution to a problem posted [here](https://github.com/gadgetron/gadgetron/issues/719#issuecomment-462853442).

version_config.cmake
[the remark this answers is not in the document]
KrisThielemans: can you point me to where the Gadgetron change is?

SuperBuild/External_Gadgetron.cmake
KrisThielemans: seems to me you want to use `find_package(blas)` all the time, and pass the variables on. People might want to pass other BLAS options. You could then do the check only on Apple.

rijobro: `find_package(blas)` creates 19 variables in total, so I think it would be unsightly to pass all of them on to Gadgetron. I'll do it if you want me to, though.

rijobro: Oh, sorry I misunderstood your comment. Ok I'll do that.

SuperBuild/External_Gadgetron.cmake
```diff
@@ -64,6 +64,15 @@ if (MKL_FOUND)
   endif ()
 endif ()
 
+  # BLAS
+  find_package(blas)
```
KrisThielemans: I was wrong. There is no `Findblas.cmake`. It turns out that there is a [FindBLAS.cmake](https://github.com/Kitware/CMake/blob/release/Modules/FindBLAS.cmake) in the making for the next CMake release I guess.

Gadgetron has its [FindCBLAS.cmake](https://github.com/gadgetron/gadgetron/blob/master/cmake/FindCBLAS.cmake). But we won't have that, unless we copy it to our own CMake directory (which I'm a bit reluctant to do, as it's a half-baked thing anyway).

So, I guess we have to delete the `find_package(blas)` line, but document that people can set these 2 variables.

Sigh.

rijobro: I think there must be another `Find*BLAS.cmake` somewhere on the system, since even on Travis (a relatively clean build), it returns: 
```
-- Found BLAS: /System/Library/Frameworks/Accelerate.framework  

-- Found CBLAS: /usr/local/Cellar/openblas/0.3.5/lib/libblas.dylib  
```
The second one is the one we want obviously.

rijobro:
Ok, can this PR be merged? Travis is finally happy for all builds (except Docker). Quite a few changes to version numbers:

Gadgetron: current `master` (gadgetron/gadgetron@b6191eaaa72ccca6c6a5fe4c0fa3319694f512ab)
ISMRMRD: `v1.4.0`
Boost (OSX): `1.68.0`
Boost (linux): `1.65.1`
GCC (linux): `GCC-6` (needed by gadgetron)
BLAS (OSX superbuild): `brew install openblas`

Regardless of `DEVEL_BUILD`=`ON`/`OFF`, versions of boost, gadgetron and ismrmrd are constant.

SuperBuild/External_Gadgetron.cmake
KrisThielemans: all the more reason to remove the `find_package(blas)` line, as it's just going to create confusion!

KrisThielemans:
I know this has been merged, but can you also update CHANGES.md to reflect this?

rijobro:
Turns out I had deleted the branch, not merged it! Merged now, with updated `CHANGES.md`.

PR #215, opened 2019-03-02
docker: separate gadgetron

### LINKED ISSUE #206: Add Travis test for using Gadgetron via Docker
As opposed to building Gadgetron, we can use its docker container. This was tried by @rijobro to fix #193. That attempt was abandoned as Travis doesn't support Docker on OSX, but we should add a test case on Linux. We should then also document how to do this (obviously by pointing to the Gadgetron doc).

casperdcl:
- [x] fix #206
- [ ] build & host our own `gadgetron` image which isn't bloated?
  + [gadgetron/gadgetron/docker/base/ubuntu_1804/Dockerfile](https://github.com/gadgetron/gadgetron/blob/master/docker/base/ubuntu_1804/Dockerfile)
  + [gadgetron/gadgetron/docker/incremental/ubuntu_1804_no_cuda/Dockerfile](https://github.com/gadgetron/gadgetron/blob/master/docker/incremental/ubuntu_1804_no_cuda/Dockerfile)
- [ ] windows support (depends on https://github.com/CCPPETMR/SIRF/issues/182)
  + `network_mode: "host"` only works on linux: https://docs.docker.com/network/host/
- [ ] provide images for both with (old) and without (this PR) `gadgetron`:
  + `sirf` and `sirf_nogadgetron` (my preference), or
  + `sirf_gadgetron` and `sirf`
- [ ] split `DOCKER_BUILD=1` into `gadgetron` and `nogadgetron`
- [ ] same with `DEVEL`
- [ ] update documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -251,7 +251,7 @@ install:
         else
           mv ~/.ccache devel/
         fi
-        $DCC build --pull core
+        $DCC build --pull core gadgetron
         # don't count the extra clone we're about to do
         # TODO: Do we need this? Is docker a distinct cloner from travis?
         # cloners_count_decrement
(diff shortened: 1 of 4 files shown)


KrisThielemans:
Abandon as we need Gadgetron libraries in the future.

PR #220, opened 2019-03-15
Ubuntu18.04

### LINKED ISSUE #127: Build Gadgetron with GPU support fails
It fails because it doesn't find one header hdf5_hl.h. 
Looking at the hdf5 build we disable [`BUILD_HDF5_HL_LIB`](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/master/SuperBuild/External_HDF5.cmake#L71)

paskino:
closes #127 
Builds a Gadgetron 3.17 and ISMRMRD 1.4


paskino:
This branch will not build on Travis as it expects to have ubuntu 18.04.

rijobro:
Hi @paskino, not sure what you're trying to do in `version_config.cmake`. I had it set up so that linux and OSX would work for the given versions of ISMRMRD and Gadgetron (regardless of `DEVEL_BUILD`). If the changes to `version_config.cmake` were unintentional, you can just `git checkout` the file. If the changes are intentional - why?

paskino:
I intentionally build Gadgetron `v3.17.0` and ISMRMRD `v1.4.0` which are the latest stable. 

Boost is just a difference in the indentation. 

What is your given version of Gadgetron and ISMRMRD?

KrisThielemans:
I guess this should be abandoned. Current master is already fine.

paskino:
Are we happy with the versions in [version_config.cmake](https://github.com/CCPPETMR/SIRF-SuperBuild/blob/17514b5c456dbae901097d3fc064fbe0e02a1e0b/version_config.cmake)?

KrisThielemans:
not with those, no, but those in https://github.com/CCPPETMR/SIRF-SuperBuild/blob/master/version_config.cmake should be fine.

PR #222, opened 2019-03-21
OSX cant use brew install boost

rijobro:


rijobro:
attempts to fix https://github.com/CCPPETMR/SIRF/issues/347

PR #223, opened 2019-03-25
Enable/disable SIRF registration from superbuild

rijobro:


rijobro:
@KrisThielemans, does this seem correct to you?

The logic is:
- If building SIRF, build SIRF's registration by default
- If building SIRF registration
    - Build NiftyReg by default
    - If not building NiftyReg, display warning (because we need NiftyReg for SIRF registration) and set building NiftyReg back to true.

If the user wants to disable NiftyReg, they have to disable SIRF registration. They can still select `USE_SYSTEM_NiftyReg=ON` if they wish.

Just need to figure out what to pass in `External_SIRF.cmake`, but this depends on the conversation here: https://github.com/CCPPETMR/SIRF/pull/353.

rijobro:
superceded by #226

PR #225, opened 2019-03-27
Fix boost osx

rijobro:
- try to get Boost for Travis OSX via miniconda.
- supercedes #222.


rijobro:
@KrisThielemans, @casperdcl - Any idea why the boost paths aren't getting passed through to Gadgetron? From the log (https://api.travis-ci.org/v3/job/511939140/log.txt), you can Travis does the following:
```
EXTRA_BUILD_FLAGS="${EXTRA_BUILD_FLAGS} -DBOOST_ROOT:PATH=/Users/travis/miniconda3/ -DBOOST_LIBRARYDIR:PATH=/Users/travis/miniconda3/lib/"
```
and yet the error during Gadgetron looks like this: 
```
n file included from /Users/travis/build/CCPPETMR/SIRF-SuperBuild/sources/Gadgetron/gadgets/spiral/vds.cpp:8:

In file included from /usr/local/include/boost/range/algorithm.hpp:64:

/usr/local/include/boost/range/algorithm/random_shuffle.hpp:33:5: error: no member named 'random_shuffle' in namespace 'std'; did you mean simply 'random_shuffle'?

    std::random_shuffle(boost::begin(rng), boost::end(rng));
```
the `/usr/local/include/boost` shows it's still picking up the brew installed version...

KrisThielemans:
very hard to convince [FindBoost.cmake](https://cmake.org/cmake/help/v3.14/module/FindBoost.html) to find what you tell it to. I'd probably try with just `BOOST_ROOT` and `-DBoost_NO_SYSTEM_PATHS=ON`. If it works, we want to add that option in the Superbuild (maybe in `External_Boost.cmake` as superbuild var but if it fails, set `BOOST_INCLUDEDIR`.

rijobro:
Sigh, neither worked.

(I set both `DBoost_NO_SYSTEM_PATHS=ON` and `BOOST_INCLUDEDIR` at the same time, would it make any difference to do them separately?)

KrisThielemans:
I don't think so.

A question for the cmake list I'm afraid.

KrisThielemans:
I don't think so.

A possibility is that `/usr/local/include` is added by something else (or even just by the compiler) and that it happens to be first in the include path. I guess you should set `Boost_DEBUG=ON` to see what it reports as include file location, or check the CMakeCache.txt. I would probably also put `VERBOSE=ON` for the Gadgetron make. All a pain to do on Travis.

If it is actually `FindBoost` picking up the wrong location, it's a question for the cmake list I'm afraid.

rijobro:
superceded by #247

PR #226, opened 2019-03-27
Update nifty reg

rijobro:
- use `NiftyRegConfig.cmake` instead of `FindNiftyReg.cmake`
- Give users the option to disable SIRF registration from Superbuild level
- NiftyReg -> NIFTYREG


KrisThielemans:
If this doesn't make it into 2.0-rc1, I think we should at least make sure that the naming of the variables will be ok.

PR #238, opened 2019-04-24
bugfix find blas

### LINKED ISSUE #233: find blas problem
we get
```
CMake Warning at SuperBuild/External_Gadgetron.cmake:68 (find_package):
  By not providing "Findblas.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "blas", but
  CMake did not find one.
```
I believe I commented about this before but didn't find it.

paskino:
closes #233

### DIFF
--- a/SuperBuild/External_Gadgetron.cmake
+++ b/SuperBuild/External_Gadgetron.cmake
@@ -65,7 +65,7 @@ if (MKL_FOUND)
 endif ()
 
   # BLAS
-  find_package(blas)
+  find_package(BLAS)
   if (APPLE AND NOT (CBLAS_LIBRARY AND CBLAS_INCLUDE_DIR))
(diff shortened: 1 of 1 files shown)


rijobro:
I can't review this unfortunately as OSX is case-insensitive (which is possibly why this slipped in in the first place).

To check that everything is OK, I would have a look at the Travis logs for the master branch and search for the warning in #233. If your fix is OK, the same warning will have disappeared in this PR's Travis log.

rijobro:
Or I suppose confirming that the warning message is no longer on your machine is good enough.

paskino:
The warning disappeared, check [here](https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/jobs/523962099#L1865) for example. You see that BLAS was found.

I'm not sure it's a Apple issue. CMake is case sensitive on both Apple and UNIX.

Issue #203, opened 2019-02-01
dependent projects need to be passed variables as cached


KrisThielemans:
We currently pass lots of directories like this
```cmake
... -DSTIR_DIR=${STIR_DIR} ...
```
This works fine. However, if you then go into the build directory of the dependent project (e.g. SIRF), and issue CMake there, it appears as if `STIR_DIR` is not set as part of the cached variables. I believe this therefore can cause problems because it might then pick up another location (only 99% sure).

The solution seems to be to do
```cmake
... -DSTIR_DIR:PATH=${STIR_DIR} ...
```
then it becomes a cached variable. Aside from less error-prone, I think it's also more user friendly. Of course, this would mean changing quite a lot of these.

Any objections?

KrisThielemans:
@rijobro can you check if adding `PATH` indeed caches it?

rijobro:
Confirmed. We should be doing this.

paskino:
Is **all** variables or just the paths? I presume it's all.

KrisThielemans:
All, indeed

PR #240, opened 2019-04-25
Add option Gadgetron_USE_MKL

### LINKED ISSUE #239: find mkl warning
as with #233 we have [this](https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/jobs/523962099#L1831) warning 

```
CMake Warning at SuperBuild/External_Gadgetron.cmake:58 (find_package):
  By not providing "FindMKL.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "MKL", but
  CMake did not find one.
```
which is originated by 
```cmake
find_package(MKL)
```

I suggest to look for MKL by setting the `BLA_VENDOR` variable.

paskino:
sets the USE_MKL variable for the configuration of Gadgetron
closes #239

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -145,6 +145,8 @@ option(USE_SYSTEM_SIRF "Build using an external version of SIRF" OFF)
 option(USE_SYSTEM_NIFTYREG "Build using an external version of NIFTYREG" OFF)
 option(USE_SYSTEM_GTest "Build using an external version of GTest" OFF)
 option(USE_SYSTEM_ACE "Build using an external version of ACE" ON)
+option(Gadgetron_USE_MKL "Instruct Gadgetron to build linking to the MKL. The user must be able to install MKL on his own." OFF)
+mark_as_advanced(Gadgetron_USE_MKL)
 
 if (WIN32)
   set(build_Gadgetron_default OFF)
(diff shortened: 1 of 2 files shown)


SuperBuild.cmake
```diff
@@ -145,6 +145,8 @@ option(USE_SYSTEM_SIRF "Build using an external version of SIRF" OFF)
 option(USE_SYSTEM_NIFTYREG "Build using an external version of NIFTYREG" OFF)
 option(USE_SYSTEM_GTest "Build using an external version of GTest" OFF)
 option(USE_SYSTEM_ACE "Build using an external version of ACE" ON)
+option(Gadgetron_USE_MKL "Instruct Gadgetron to build linking to the MKL. The user must be able to install MKL on his own." OFF)
```
KrisThielemans: move these to External_Gadgetron.cmake.

paskino: If I put the option in `External_Gadgetron.cmake` it will not possible to build Gadgetron linking to MKL from the SuperBuild. The user should then modify `External_Gadgetron.cmake`. 

Or do I miss something @KrisThielemans ?

KrisThielemans: We have STIR specific options in `External_STIR.cmake`. I cannot see why this wouldn't work here.

paskino:
Two comments:

1. [it seems](https://github.com/Kitware/CMake/blob/master/Modules/FindBLAS.cmake#L142L148) that one should set the environment variable `LD_LIBRARY_PATH` to point to the location of the MKL to get it found by `find_package(BLAS)` 
2. [Gadgetron seems](https://github.com/gadgetron/gadgetron/blob/master/CMakeLists.txt#L348) to disable `USE_MKL` by default. So passing `Gadgetron_USE_MKL` doesn't work.

KrisThielemans:
> one should set the environment variable LD_LIBRARY_PATH...

that doesn't surprise me. It'll probably do a test if some test code using mkl compiles and runs. That is not an issue for CMake, but for you local config. (i.e. your `ldconfig` set-up or `LD_LIBRARY_PATH` needs to contain what you need)

> Gadgetron seems to disable USE_MKL by default. 

no, it creates an option `USE_MKL` with default `OFF`. You might have to pass `USE_MKL:BOOL=...` instead. See also #203

paskino:
yes, 1. was a reminder for me, and maybe we can add the info to the wiki.

PR #251, opened 2019-06-21
HDF5 updates for Windows

KrisThielemans:
resolve issues with MATLAB compatibility. With this update, it is now recommended to leave `USE_SYSTEM_HDF5=OFF` on Windows.


KrisThielemans:
@evgueni-ovtchinnikov could you please try this? Windows SuperBuild should now work with boost preinstalled. It should work for MATLAB, or for Python (if you set `BUILD_MATLAB=OFF`).

Of course, I'd recommend to build in a different directory such that it doesn't interfere with your normal set-up.

evgueni-ovtchinnikov:
tried to build new SIRF-SuperBuild (I understand it is on master now that you merged WINHDF5 there), succeeded after removing the contents of old INSTALL/cmake (otherwise failed with weird error messages)

could not find `BUILD_MATLAB` in CMake though, so not entirely sure I build the right SIRF-SuperBuild

evgueni-ovtchinnikov:
@rijobro: SIRF would not build complaining about `fabs` on line 750 of `NiftyImageData.cpp`, so I had to cast the argument into float:
```
    if(float(fabs(float(elem1-elem2))) < 1.e-7F)
```
ok do you think?

SIRF build failure was rather weird: complained about unresolved setParameter, but building `Reg` only succeeded (after fixing `fabs`) as well as subsequent builds of the whole SIRF!

Python and Matlab REG_TESTs failed though

evgueni-ovtchinnikov:
the reason for `REG_TEST_PYTHON` failure was that I forgot to update `PYTHONPATH`, after update it works, but `MR_TESTS_PYTHON` now fails in Visual Studio (ok from command line)

evgueni-ovtchinnikov:
the reason for `REG_TEST_MATLAB` failure was that function definitions in scripts are not allowed in MATLAB R2015b that I have on my desktop

rijobro:
> `if(float(fabs(float(elem1-elem2))) < 1.e-7F)`

This is ok for me

> Including functions in scripts requires MATLAB R2016b or later.

I've just seen this on the Matlab website. I'll have to change the ctests so that they no longer use functions

rijobro:
PR https://github.com/CCPPETMR/SIRF/pull/405 should solve your Matlab problem.

PR #252, opened 2019-06-23
added option to disable Python or Matlab support

KrisThielemans:
This can be used if do have Python/Matlab on your system but don't want to include support for it. This is useful if you have compilation or run-time errors  with one of them (e.g. because of mismatches shared libraries) and you don't need it.

Currently only passed to SIRF. Option names are `DISABLE_Python` etc, maybe better `SIRF_`DISABLE_Python` ?


SuperBuild.cmake
```diff
@@ -70,63 +70,79 @@ mark_as_superbuild(
...
-if (PYTHONLIBS_FOUND)
-  message(STATUS "Found PYTHON_INCLUDE_DIRS=${PYTHON_INCLUDE_DIRS}")
-  message(STATUS "Found PYTHON_LIBRARIES=${PYTHON_LIBRARIES}")
-endif()
+#### Python support
+
+option(DISABLE_PYTHON "Disable building SIRF python support" OFF)
+if (DISABLE_PYTHON)
+  message(STATUS "Python support disabled")
+else(DISABLE_PYTHON)
```
paskino: `DISABLE_PYTHON` in `else(DISABLE_PYTHON)` is not needed and possibly confusing.

paskino:
The code seems fine to me, though I haven't tested it. 

w.r.t. naming, it doesn't seem we have a `SIRF_` policy on variable names passed to SIRF compilation. Does it make sense to add `SIRF_DISABLE_Python` or it is just very verbose?

KrisThielemans:
> it doesn't seem we have a SIRF_ policy on variable names passed to SIRF compilation.

I think we do...

can you please rename to `SIRF_DISABLE...` then? We can always have a general `DISABLE` then later on

SuperBuild.cmake
KrisThielemans: I guess I was following other style from somewhere else in the file, but I agree. (These days CMake doesn't recommend it anymore either). Please remove it from the `endif()` as well then

KrisThielemans:
I think we discussed that this functionality was missing. Anyone volunteering to fix this up and merge?

KrisThielemans:
I've  updated this. Will probably merge soon. If anyone with with both MATLAB and Python wants to try, that'd be good.

Let' s postpone introducing `SIRF_DISABLE_Python` variables until we make a macro for it such that it will be easier to do this for all projects.

PR #264, opened 2019-09-25
upgrade Gadgetron

rijobro:

PR #266, opened 2019-09-25
try to fix osx travis by building own version of boost

rijobro:


KrisThielemans:
Please check #264 first

rijobro:
I'd rather go with #264, which uses an updated version of Gadgetron. Commit https://github.com/gadgetron/gadgetron/commit/8ea7f911e87575f666e30bfa5ffe7269023b0480 removes dependence on `boost::thread`, which is the source of our problems.

Unfortunately, I can't get new versions of Gadgetron installed (not invested much time in it). Therefore I thought it might be easiest to simply keep the old version of Gadgetron and install boost ourselves (instead of `brew install boost`). This seems like the easy solution providing the builds don't time out/max out the log length.

PR #272, opened 2019-10-14
Build STIR with NiftyPET

rijobro:
Use STIR with NiftyPET projectors if:

- CUDA is present
- The superbuild is building STIR


version_config.cmake
```diff
@@ -138,6 +138,10 @@ set(DEFAULT_astra-toolbox_TAG v1.8.3)
 set(DEFAULT_TomoPhantom_URL https://github.com/dkazanc/TomoPhantom )
 set(DEFAULT_TomoPhantom_TAG v1.4)
 
+## NiftyPET
+set(DEFAULT_NIFTYPET_URL https://github.com/pjmark/NIPET )
+set(DEFAULT_NIFTYPET_TAG 70b97da0a4eea9445e34831f7393947a37bc77e7)
```
casperdcl: might want to check with @pjmark if this commit is fine

rijobro: You're right, just went with most recent commit. @pjmark ?

PR #278, opened 2019-10-24
CCPi patch script

paskino:
Adds a patch script for the CCPi-Framework installation.


patches/cil-patch.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import print_function, division
```
KrisThielemans: I guess you don't need this line

paskino: ah! you know, it's safe anyway.

KrisThielemans:
restarted the last Travis job. network problems

paskino:
of course the patch fails on the VM with git paranoia: the patch changes the file in the repo under git control

paskino:
Works now.

Issue #167, opened 2018-12-08
travis 4MB log size limit


KrisThielemans:
we're now very close to the log file limit. Travis aborts a job if it goes over. This currently happens in #166 for one [OSX job](https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/jobs/465422525).

It doesn't seem to be possible to raise this limit, see e.g. https://github.com/travis-ci/travis-ci/issues/1382. We could try to save logs to a file, but
- you'd need to see if what went wrong, so would maybe want to upload it somewhere, which [could be AWD](https://blog.travis-ci.com/2012-12-18-travis-artifacts) with extra effort.
- travis will then time-out, which could be solved via a `travis_wait`

See [here](https://stackoverflow.com/questions/26082444/how-to-work-around-travis-cis-4mb-output-limit/26082445#26082445) for work-around (that makes the travis file more complicated...) Note that this trick would need work when using parallel builds.

sigh

KrisThielemans:
This issue, together with the fact that building boost 1.68.0 (and earlier) issues a ton of warnings when building with Clang like
```
./boost/preprocessor/tuple/elem.hpp:40:37: warning: variadic macros are a C99 feature [-Wvariadic-macros]
```
prevents us to test `-DUSE_SYSTEM_Boost=OFF -DUSE_SYSTEM_HDF5=OFF`.  Oh well.

KrisThielemans:
we won't update Travis things anymore

PR #284, opened 2019-11-04
travis & docker overhaul

### LINKED ISSUE #268: Travis updates
- use stages (https://docs.travis-ci.com/user/build-stages/)
    + use conditionals for deployment, maybe allow Mac builds to fail https://docs.travis-ci.com/user/conditional-builds-stages-jobs/
- may help with #167
- change default `dist` to 16.04 Xenial (https://docs.travis-ci.com/user/trusty-to-xenial-migration-guide/) or even 18.04 Bionic (https://docs.travis-ci.com/user/reference/bionic/) instead of Trusty 14.04 (but perhaps retain one trusty build)

### LINKED ISSUE #281: fix docker ccache
Current builds (https://travis-ci.org/CCPPETMR/SIRF-SuperBuild) take about 4 times longer on `docker` because `.ccache` is no longer being used properly. Needs to be updated/re-fixed.

Also we should be re-building without caching `ccppetmr/sirf:core` (and also maybe not caching `ccppetmr/sirf:base`)

casperdcl:
- [x] fixes #281 docker caching
  + [x] revisit docker image caching
  + [x] refix docker `ccache`
- [x] fixes #268 travis upgrade
  + [x] use build stages
  + [x] use conditionals for deployment
  + [x] allow Mac builds to fail
  + [x] change default to `dist: bionic` (18.04)
  + [x] add `dist: xenial` (16.04)
  + [x] keep one `dist: trusty` (14.04)
  + [x] restructure jobs matrix to avoid duplication and unnecessary commands
- may help with #167


casperdcl:
[diff without whitespace](https://github.com/CCPPETMR/SIRF-SuperBuild/pull/284/files?diff=unified&w=1)

docker/user_sirf-ubuntu.sh
```diff
@@ -3,11 +3,15 @@
 set -ev
 INSTALL_DIR="${1:-/opt}"
 # SIRF
-git clone https://github.com/CCPPETMR/SIRF-SuperBuild --recursive -b master $INSTALL_DIR/SIRF-SuperBuild
+git clone https://github.com/CCPPETMR/SIRF-SuperBuild --recursive -b docker-ccache $INSTALL_DIR/SIRF-SuperBuild
```
KrisThielemans: seems incorrect

.travis.yml
```diff
@@ -1,143 +1,321 @@
...
-sudo: required  # for docker
-services:
-- docker
+dist: bionic  # Ubuntu 18.04 LTS
 cache:  # shared between builds
 - ccache
 - pip
+stages:
+# - test  # TODO: uncomment before merge
+- name: test  # TODO: remove before merge
```
KrisThielemans: I guess the `TODO` should have been removed as well (also below)

docker/user_sirf-ubuntu.sh
casperdcl: yes I pushed a commit to master immediately after merge reverting this. Just want to force refresh master trvis caches. Will take a while to build but everything (except of course osx) should pass.

.travis.yml
casperdcl: yes, was part of merge commit

casperdcl:
hmm before there's: https://github.com/CCPPETMR/SIRF-SuperBuild/blob/ea19bf415a2e7b00696752ee751859d65ca85f17/SuperBuild.cmake#L61 and after: https://github.com/CCPPETMR/SIRF-SuperBuild/blob/ea19bf415a2e7b00696752ee751859d65ca85f17/SuperBuild.cmake#L262 is that correct?
Will try reverting these changes later in another branch once all the caches have finished rebuilding and see if it makes a difference.

KrisThielemans:
https://github.com/CCPPETMR/SIRF-SuperBuild/blob/ea19bf415a2e7b00696752ee751859d65ca85f17/SuperBuild.cmake#L61
not sure what that does, but I believe @rijobro had to put it in for the library names. can't see what would be wrong with it.

https://github.com/CCPPETMR/SIRF-SuperBuild/blob/ea19bf415a2e7b00696752ee751859d65ca85f17/SuperBuild.cmake#L262
> is that correct?

why wouldn't it be?  somewhat weirdly, all projects currently depend on `PRIMARY_PROJECT_NAME`, but that doesn't have anything to do with `mark_as_superbuild`

PR #288, opened 2019-11-12
FIX: renaming petmr_rd_tools to pet_rd_tools was incomplete

ashgillman:
This changes all petmr_rd_tools references to pet_rd_tools.


ashgillman:
Also there was an issue with the Glog tag.

ashgillman:
Just a link for the wiki, it still worked but I've updated it

KrisThielemans:
I guess none of this is over tested in Travis. would it make sense to add it to at least one of the docker builds for instance? @casperdcl ?

ashgillman:
It might make sence to at least check that `nm_extract` runs without returning an error. That wouldn't require DICOM test data, and at least makes sure that the binary is linked against ITK etc.

ashgillman:
I couldn't find any references to the tools in the User Guide.

KrisThielemans:
>It might make sence to at least check that nm_extract runs without returning an error. 

even just building would be fine. I'm fairly confident that if it builds, it'd work (as long as run-time library paths would be set ok)

ashgillman:
I'll defer to @casperdcl , I had a look at .travis.yml and I'm not sure how that thing is working. Does just adding "+itk" to the name make it build ITK?

ashgillman:
I've also updated the PET RD Tools default to the latest stable version.

KrisThielemans:
@ashgillman please add to `CHANGES.md` such that we can merge this

ashgillman:
Done

PR #299, opened 2019-11-19
disabled hdf5 by default in external_stir.cmake

ALEXJAZZ008008:
There was an issue, for me, where I could not use SIRF as it would throw a HDF5 related error whenever I tried to do anything with it. I have disabled HDF5 by default in External_STIR.cmake which solves the problem for me.


rijobro:
Can you give more information on the error? I'm not sure how HDF5 is used in STIR but it seems a shame to disable it if we can solve the problem.

rijobro:
Also, although it's not merged yet, https://github.com/CCPPETMR/SIRF-SuperBuild/pull/289 will mean that STIR variables are passed from the SuperBuild as cached variables. This would mean that your addition would have to be given as `-DDISABLE_HDF5_SUPPORT:BOOL=ON`. So if we accept this PR, that will have to be changed :)

ALEXJAZZ008008:
Whenever I tried to do quite literally anything (either using the STIR binaries or importing SIRF in Python) I would get this error "terminate called after throwing an instance of 'H5::PropListIException'". I didn't try debugging it in a C++ IDE but I imagine that might help find the issue.
I have amended the PR as you asked

rijobro:
Ok, how about in `External_STIR.cmake` you create an option along the lines of `STIR_DISABLE_HDF5_SUPPORT`. Default to `OFF`. Then pass to STIR with:
```
-DDISABLE_HDF5_SUPPORT:BOOL=${STIR_DISABLE_HDF5_SUPPORT}
```
That way, for most users functionality will be unchanged, but you'll have the option to disable support.

ALEXJAZZ008008:
I've made the change you suggested. The only reason I wrote it as I did in the first place is because I was told that it was expected for HDF5 support to be turned off when STIR was built using the superbuild, it can be on or off now.

rijobro:
Ok, I have no strong opinion as to what the default should be, I'll leave that for @KrisThielemans. The rest looks good for me.

KrisThielemans:
This is fine for me, although we'll need to switch the default to `ON` once STIR Signa support has been merged (and we're using it in SIRF).

At present, STIR `master` nor 4.0.0-alpha doesn't use HDF5, so let's postpone merging this after tagging the SB for v2.1.

KrisThielemans:
Note that the run-time error seenby @ALEXJAZZ008008 was most likely related to 2 different versions of HDF5 used for compilation/running.

PR #316, opened 2020-02-14
spm12

rijobro:


KrisThielemans:
By the way, there isn't anything specific to spm12 here as far as I can see.

rijobro:
What do you mean nothing specific to spm12? There's `FindSPM12.cmake` and `External_SPM12.cmake`.

KrisThielemans:
I meant version 12. `spm_realign` is quite old. probably was in SPM2
https://www.fil.ion.ucl.ac.uk/spm/software/

Can you add in the README for this that SPM is licensed under GPL2, and that implies that any functionality that uses this SPM wrapper  needs to be distributed under GPL2 as well

rijobro:
Ah, I referred to it as SPM12 because each of the versions are in a github repository of their own: https://github.com/spm/. 

What do you think I should do? I'm in favour of only supporting spm12 as it reduces our amount of work; are that many people using older versions, anyway?

rijobro:
Both `spm_realign.m` and `spm_reslice.m` are present in `spm99`, so I guess the `FindSPM12.cmake` could be renamed to `FindSPM.cmake` and returns true provided those two files are present. 

I've got absolutely no idea how backwards compatible `spm_realign.m` and `spm_reslice.m` are over time.

KrisThielemans:
I certainly don't want to support older version s of SPM (although I'm 99% sure that nothing changed in that code).

In any case, we're currently not checking version etc, so I suggest to rename to `FindSPM.cmake` and just add in the doc that it has only been tested with SPM12.

PR #317, opened 2020-02-17
202001 hackathon

paskino:
Builds versions of SIRF and CIL that enable MCIR. 
should close #310


KrisThielemans:
@paskino this might need your attention now

paskino:
I believe current master and #368 supersede this PR

PR #321, opened 2020-02-18
change FFTW_URL to ftp source

### LINKED ISSUE #320: FFTW has removed version 3.3.5 from http site
Can't download and find version 3.3.5 now and build will fail.

Should increase the version in `version_config.cmake`

paskino:
closes #320

### DIFF
--- a/version_config.cmake
+++ b/version_config.cmake
@@ -55,7 +55,7 @@ if(WIN32)
   #set(FFTW3_URL https://s3.amazonaws.com/install-gadgetron-vs2013/Dependencies/FFTW/zip/FFTW3.zip )
   #set(FFTW3_MD5 a42eac92d9ad06d7c53fb82b09df2b6e )
 else(WIN32)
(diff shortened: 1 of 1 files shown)


paskino:
I tested locally and it works, however travis fails to get the fftw file from ftp. Any ideas?

KrisThielemans:
very long time ago we had the same problems I belief. Their `ftp` server is probably not very stable. Or Travis blocks it. Or whatever... stick to `http` I guess

paskino:
Ok, then I Will need to change fftw version as they removed 3.3.5 from http.

KrisThielemans:
Can put it elsewhere, e.g. or website. You've done that before

paskino:
it seems the file is back on [http://www.fftw.org/fftw-3.3.5.tar.gz](http://www.fftw.org/fftw-3.3.5.tar.gz) so I will close the PR.

PR #335, opened 2020-02-27
make possible to disable NIFTYREG checkouts

AnderBiguri:
Allows to disable the Checkouts for Niftyreg. 
Also passes, if set, the variable CMAKE_POSITION_INDEPENDENT_CODE, to be used by NIFTYREG to pass the CMAKE_C_COMPILE_OPTIONS_PIC string, so we can flag "-fPIC" in the build of NIFTYREG

See: https://github.com/KCL-BMEIS/niftyreg/pull/67


AnderBiguri:
@KrisThielemans Done!

AnderBiguri:
I have made a big mistake from the beginning with this PR: I made it from my master branch and then fiddled with it. I have tried to solve it but history is too entangled to fully fix it and this will create a messy past in the repo. I will therefore just delete this PR, and open a new clean one.

PR #339, opened 2020-02-28
Add flag for position dependent library compiling 

AnderBiguri:
Otherwise NiftyReg won't pass -fPIC to the CUDA files, thus will complain.

This requires the following to work: KCL-BMEIS/niftyreg#67


AnderBiguri:
Hum, Travis doesn't like my changes. I don not know why/how to fix it. Happy to take suggestions.

rijobro:
Looks good to me, one of our docker builds is broken. See here: https://github.com/CCPPETMR/SIRF-SuperBuild/issues/331

PR #344, opened 2020-03-03
Fix 343

### LINKED ISSUE #343: SPM path is not passed to MATLABPATH
I have this problem where the `ctest` relating to SPM in MATLAB fail because the path of SPM is not known to MATLABPATH. 

I have a solution, will PR soon.

AnderBiguri:
Resolves #343. There is no need of a conditional statement, as MATLAB will only accept existing paths and will ignore any other paths/text/arguments that do not exist. if SMP_DIR does not exist, this still works.

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -318,7 +318,7 @@ set(ENV_MATLAB_BASH "#####     Matlab not found     #####")
 set(ENV_MATLAB_CSH  "#####     Matlab not found     #####")
 if (Matlab_FOUND)
   set(ENV_MATLAB_BASH "\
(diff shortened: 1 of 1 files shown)


rijobro:
Looks fine to me. Did you test this when Matlab is present, but SPM isn't? Travis doesn't have Matlab support so we need to be extra sure before merging any Matlab changes!

AnderBiguri:
I have tested it when MATLAB is there but not SPM, and by adding various real and fake paths interleaved and MATLAB will only take whichever paths exist.

rijobro:
ok. No point waiting for Travis as the code won't be tested.

PR #351, opened 2020-03-09
Clean up conditional git CMAKE args

### LINKED ISSUE #336: clean-up conditional git checkout code
It would be shorter/cleaner to set the `git` arguments to a macro, like I did with the directory arguments #326. I suggest to create a variable `${proj}_EP_ARGS_GIT` in `SetGitTagAndRepo`  to replace the usage of
```cmake
GIT_REPOSITORY "${${proj}_URL_MODIFIED}"
GIT_TAG "${${proj}_TAG_MODIFIED}"
```
(or equivalent without `_MODIFIED`) in the `External_*` cmake files.

Seems that we could then replace https://github.com/CCPPETMR/SIRF-SuperBuild/blob/602922caf315b073b814aa1c3f8806f2f50e4db9/CMake/SetGitTagAndRepo.cmake#L39-L48 with
```cmake
         # set a variable to use in the ExternalProject_Add call that sets all git-related variables
	if (EXISTS "${${proj}_SOURCE_DIR}" AND DISABLE_GIT_CHECKOUT_${proj})
                set(${proj}_EP_ARGS_GIT)
		message(STATUS "Not going to perform git checkout for ${proj}...")
	else()
		message(STATUS "Will perform git checkout for ${proj}...")
                set(${proj}_EP_ARGS_GIT
                    GIT_REPOSITORY ${${proj}_URL}
                    GIT_TAG ${${proj}_TAG}
                 )
             endif()
	endif()
```

AnderBiguri:
Resolves #336

I can add this too to PR #350 if that gets merged. 
However this still has the issue that @rijobro mentioned: In order to change the git checkout, you need to have the repo downloaded already, otherwise it will not work. I do not know how to fix that, I am open to suggestions.

### DIFF
--- a/CMake/SetGitTagAndRepo.cmake
+++ b/CMake/SetGitTagAndRepo.cmake
@@ -38,12 +38,13 @@ macro(SetGitTagAndRepo proj)
 	# If disable git desired (and source directory exists), set tag and repo to blank
 	if (EXISTS "${${proj}_SOURCE_DIR}" AND DISABLE_GIT_CHECKOUT_${proj})
 		message(STATUS "Not going to perform git checkout for ${proj}...")
-		SET(${proj}_TAG_MODIFIED "")
-		SET(${proj}_URL_MODIFIED "")
+		set(${proj}_EP_ARGS_GIT "")
 	# Else, 
 	else()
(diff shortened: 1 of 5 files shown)


AnderBiguri:
@KrisThielemans  updated

PR #357, opened 2020-03-15
[TRAVIS] OSX: find latest openblas version

### LINKED ISSUE #356: Travis OSX failure with BLAS
https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/jobs/662503202 failed, which is the Python3 job for OSX. https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/jobs/662503202#L15192 says
```
dyld: Library not loaded: /usr/local/opt/openblas/lib/libopenblasp-r0.3.7.dylib
  Referenced from: /Users/travis/build/CCPPETMR/SIRF-SuperBuild/builds/Gadgetron/build/test/test_all
  Reason: image not found
```
I think the following is going on:
1. `brew install openblas` says 
  ```
  openblas 0.3.7 is already installed
  To upgrade to 0.3.9, run `brew upgrade openblas`
  ```
2. a bit further on, the log file says
  ```
  `brew cleanup` has not been run in 30 days, running now
 .  ......
   Removing: /Users/travis/Library/Caches/Homebrew/openblas--0.3.7.mojave.bottle.tar.gz
   .......
   Installing postgis dependency: openblas 0.3.9
  ```

So it's likely that our work-around to find the correct blas files
https://github.com/CCPPETMR/SIRF-SuperBuild/blob/01a15530fcdbc90204e01719300e04346d7d6ec4/.travis.yml#L377-L379
fails, as that happens between step 1 and 2, i.e. it will point to 0.3.7 which is no longer there.

KrisThielemans:
`brew` can change versions on packages due to dependencies etc. So we need to find the BLAS version
just before starting to build.

Fixes #356

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -373,7 +373,7 @@ before_install:
      brew update
      BUILD_FLAGS="$BUILD_FLAGS -DSHARED_LIBS_ABS_PATH=ON"
      brew install openblas
-     # Let CMake find this openblas version (note: we use /usr/local/opt/openblas symlink to get the most recent brew version)
+     # Let CMake find this blas version (note: we use /usr/local/opt/openblas symlink to get the most recent brew version)
      EXTRA_BUILD_FLAGS="$EXTRA_BUILD_FLAGS -DCBLAS_INCLUDE_DIR=/usr/local/opt/openblas/include -DCBLAS_LIBRARY=/usr/local/opt/openblas/lib/libblas.dylib"
      if [ $PYMVER == 2 ]; then
        PYINST=/System/Library/Frameworks/Python.framework/Versions/$PYMVER.7
@@ -422,7 +422,6 @@ before_install:
        #export PATH="$PWD/cmake/bin:$PATH"
      )
    elif [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
-     exit 1 # KTTRAVIS
      PY_EXE=python$PYMVER
      if [[ -z "$DOCKER_BUILD" ]]; then
        curl -o cmake.tgz -L https://github.com/Kitware/CMake/releases/download/v3.13.4/cmake-3.13.4-Linux-x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


KrisThielemans:
If this works, we'll need similar changes in SIRF, and  indeed our installation instructions.

KrisThielemans:
oh well. [one job](https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/jobs/662566361?utm_medium=notification&utm_source=github_status) fails with
```
.../SIRF-SuperBuild/sources/Gadgetron/toolboxes/core/cpu/math/cpp_blas.cpp:7:18: fatal error: 'cblas.h' file not found
```

KrisThielemans:
no travis job scheduled for some reason. I'll close and reopen

KrisThielemans:
By the way, `brew` says
```
==> Pouring openblas-0.3.9.mojave.bottle.tar.gz
==> Caveats
openblas is keg-only, which means it was not symlinked into /usr/local,
because macOS provides BLAS and LAPACK in the Accelerate framework.
For compilers to find openblas you may need to set:
  export LDFLAGS="-L/usr/local/opt/openblas/lib"
  export CPPFLAGS="-I/usr/local/opt/openblas/include"
For pkg-config to find openblas you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/openblas/lib/pkgconfig"
```
So, are we sure we need this version specific code? https://github.com/CCPPETMR/SIRF-SuperBuild/blob/01a15530fcdbc90204e01719300e04346d7d6ec4/.travis.yml#L377-L379

KrisThielemans:
The current strategy fails as `brew list --versions openblas`  returns more than 1 version and we get
`-DCBLAS_INCLUDE_DIR=/usr/local/Cellar/openblas/0.3.7 0.3.9/include -DCBLAS_LIBRARY=/usr/local/Cellar/openblas/0.3.7 0.3.9/lib/libblas.dylib`

KrisThielemans:
This now works fine. @rijobro could you make the equivalent change in SIRF and adjust install instructions (wiki I guess).

let's squash merge this.

Issue #346, opened 2020-03-05
External_CCPi-Framework doesn't install libcilacc.so


rijobro:
As far as I can see, `External_CCPi-Framework.cmake` modifies the install command:

https://github.com/CCPPETMR/SIRF-SuperBuild/blob/d5a2e679579b149b910e4cbff2ad908f94771d48/SuperBuild/External_CCPi-Framework.cmake#L68

This means that `libcilacc.so` isn't installed, leading to my current error: 
```
OSError: libcilacc.so: cannot open shared object file: No such file or directory
```

rijobro:
Actually i don't understand as the build step is also: `BUILD_COMMAND ""`. So how will `libcilacc.so` ever be compiled?

rijobro:
Also related: https://github.com/vais-ral/CCPi-Framework/issues/505

PR #359, opened 2020-03-16
Fix cil

rijobro:
This PR should mean that libraries are compiled/installed and that python wrappers are installed into correct location


rijobro:
fixes #346

Issue #257, opened 2019-07-25
re-running cmake to change tags (or DEVEL_BUILD=on) doesn't work


KrisThielemans:
@evgueni-ovtchinnikov and I tried the following:

1. build with default options (i.e. latest release)
2. `cmake -USTIR_TAG -USIRF_TAG -DDEVEL_BUILD=ON .; make`
  This did not update the tags to `master` in the STIR and SIRF clones.
3. `cmake -STIR_TAG=master -DSIRF_TAG=master .; make`
  This didn't help either.

Note that the tags in the Superbuild's `CMakeCache.txt` were fine. It's just that `ExternalProject` stuff doesn't seem to do what we think it does.

The only thing that worked was
```
rm -rf sources/STIR sources/SIRF builds/STIR builds/SIRF
```

PR #361, opened 2020-03-18
update CIL framework to point to current master

rijobro:


rijobro:
Both this PR and the master seem to be in the same state. Travis is happy for all builds except Docker. The Docker errors look unrelated to this PR.

KrisThielemans:
These are the same jobs that were failing earlier since your previous PR, which flipped the situation between the 3rd job failing, to 3 others failing. 

I think the docker builds are the currently the only ones that try to build CIL (as @rijobro disabled the direct tests).

https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/jobs/663922430#L1584
still says
```
CMake Error: The source directory "/opt/SIRF-SuperBuild/sources/CCPi-Framework" does not appear to contain CMakeLists.txt.
```

rijobro:
It's hard to say what's going on as the output of different dependencies are overlapping. But it looks like it's getting old CCPi-Framework commit (`9bd69b7 fix MixedL21Norm (#417)`, [this line](https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/jobs/663922430#L1576)).

Is it possible that Docker uses the tags differently? Or caches them? Otherwise, I don't really see where I could have gone wrong (desired commit was `adf4163`).

rijobro:
> @rijobro please add this TAG for CCPi-Astra 8c2493710f5eac1316945585fe6333ab56f9a2a1

Done.

Presumably, it still doesn't solve the Docker issue though. @KrisThielemans , @casperdcl the Docker must be working on a different logic for checking out branches?

KrisThielemans:
I vaguely recall that the docker image just gets updated (as opposed to created from scratch at every build), which would make a lot of sense. So we might be hitting something like #257

rijobro:
I vaguely recall Docker only updating once something was in master? I'll merge this and see if it solves that problem.

casperdcl:
If you want to test docker in PRs you'll need to temporarily update https://github.com/CCPPETMR/SIRF-SuperBuild/blob/c5b52b21661d5997a76528124c317eadd5ce21ec/docker/user_sirf-ubuntu.sh#L6

PR #365, opened 2020-03-27
allow disabling cuda

rijobro:


rijobro:
@KrisThielemans any idea why travis has disappeared?

KrisThielemans:
not sure. we had it in STIR as well.
https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/requests shows it's accepted. Build is received https://travis-ci.org/github/CCPPETMR/SIRF-SuperBuild/builds/667806754. nothing running yet.

so seems to be that github doesn't get told about it by Travis (or the other way around). Probably something that'll disappear, but could check Travis issues

SuperBuild.cmake
```diff
@@ -242,11 +245,13 @@ if (USE_ITK)
 endif()
 
 # If building STIR and CUDA present, offer to build NiftyPET
-if (CUDA_FOUND AND NOT USE_SYSTEM_STIR)
+if (CUDA_FOUND AND NOT DISABLE_CUDA AND NOT USE_SYSTEM_STIR)
```
KrisThielemans: seems to me we don't need this change, as `CUDA_FOUND` won't be set. It's harmless of course, but  more verbose

SuperBuild/External_CCPi-Regularisation-Toolkit.cmake
```diff
@@ -79,6 +79,12 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
     # Sets ${proj}_URL_MODIFIED and ${proj}_TAG_MODIFIED
     SetGitTagAndRepo("${proj}")
 
+    if (CUDA_FOUND AND NOT DISABLE_CUDA)
```
KrisThielemans:
```suggestion
    if (CUDA_FOUND)
```

rijobro: I wonder if you ran CMake once with `DISABLE_CUDA=OFF`, then `CUDA_FOUND` might be set to true. If you then re-run with `DISABLE_CUDA=ON`, what will the value of `CUDA_FOUND` be? Would it stay `TRUE`? I wasn't sure so I put in the `AND` statement.

KrisThielemans: True! Suggested solution:
```cmake
if (DISABLE_CUDA)
  set(CUDA_FOUND OFF)
else()
...
endif()
```

KrisThielemans: looks like we should use that paradigm for other `DISABLE` options really. want to raise an issue if we don't?

paskino: Doesn't this require that the external projects allow passing a variable to disable CUDA? I'm pretty sure the regularisation toolkit doesn't allow you that currently.

rijobro: Good point, we should add them where possible. I was using this to disable NiftyPET. We can add in the options to the external projects as and when we need to?

rijobro: Turns out the regularisation kit does allow for disabling CUDA.

SuperBuild.cmake
rijobro: In the end, I went with `USE_CUDA = CUDA FOUND AND NOT DISABLE_CUDA`, and then only use `USE_CUDA` throughout the code.

Issue #309, opened 2019-12-16
CIL currently failing on Travis


rijobro:
PR #307 only changes a few lines of code, but many of the Travis builds are failing with errors along the lines of:
```
======================================================================
ERROR: test_DataProcessor (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_DataProcessor
Traceback (most recent call last):
  File "/usr/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/build/CCPPETMR/SIRF-SuperBuild/sources/CCPi-Framework/Wrappers/Python/test/test_DataProcessor.py", line 21, in <module>
    from ccpi.framework import DataProcessor
  File "/home/travis/build/CCPPETMR/SIRF-SuperBuild/INSTALL/python/ccpi/framework/__init__.py", line 30, in <module>
    from .framework import DataContainer
  File "/home/travis/build/CCPPETMR/SIRF-SuperBuild/INSTALL/python/ccpi/framework/framework.py", line 47, in <module>
    cilacc = ctypes.cdll.LoadLibrary(dll)
  File "/usr/lib/python3.5/ctypes/__init__.py", line 425, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python3.5/ctypes/__init__.py", line 347, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libcilacc.so: cannot open shared object file: No such file or directory
```
I haven't done any digging, but I suspec this is due to an upstream change to CIL?

rijobro:
PR #310 will temporarily disable Travis CIL until fixed.

paskino:
Will be closed by #368

PR #368, opened 2020-03-29
Cil devel build tag

### LINKED ISSUE #304: CIL_VERSION undefined for DEVEL_BUILD
https://github.com/CCPPETMR/SIRF-SuperBuild/blob/5136e69a153db16939ef67081ef85c2c30cf493e/version_config.cmake#L213

`CIL_VERSION` is undefined for `DEVEL_BUILD=ON`, which means that CIL packages can't be built (CCPi-Regularisation requires it as an environmental variable, for example).

paskino:
closes #304 by removing CIL repos from the `DEVEL_BUILD`. This is due to backward incompatibility of the structure of the CIL package from version 19.10 on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,8 +83,8 @@ jobs:
       - *addons_apt_packages
       - g++-6
    python: 3.6
-   name: py36 gcc6 +xenial -boost +fftw3 +hdf5 +ace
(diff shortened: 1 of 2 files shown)


paskino:
closes #309

rijobro:
@paskino, if I'm not mistaken this PR does two things: 

1. Add extra CIL builds to the Travis matrices
2. Removes code duplication by putting CPPi tags and URLs outside of the `if DEVEL_BUILD...ELSE` logic, since they were the same anyway.

Is that correct? If so, I'm fine with this.

paskino:
That's right.

Issue #382, opened 2020-04-29
STIR project dependencies have to be set after options


KrisThielemans:
we're setting the dependencies based on values of options
https://github.com/SyneRBI/SIRF-SuperBuild/blob/e10fe7bd8c8f28bef06d8eff386dc7b94570bf93/SuperBuild/External_STIR.cmake#L27-L29
which are however only defined later on. That means that
- there will be 2 CMake variables (the cached and non-cached) with conflicting values
- the dependencies are incorrect

KrisThielemans:
@rijobro would you be able to have a look at this in absence of @paskino ? Might also occur in other projects.

rijobro:
mv lines 25-39 in `External_STIR.cmake` after last `option`.

PR #385, opened 2020-05-01
need all options before adding dependencies

rijobro:
fixes https://github.com/SyneRBI/SIRF-SuperBuild/issues/382


KrisThielemans:
one Travis failure but Gadgetron time-out

Issue #301, opened 2019-11-20
CMAKE_BUILD_TYPE is not passed through from SuperBuild to 'External'


ALEXJAZZ008008:
The only 'External' that CMAKE_BUILD_TYPE is passed through to (it appears) is ITK.

KrisThielemans:
Need to `mark_as_superbuild`and remove the `EXTERNAL_PROJECT_BUILD_TYPE`

KrisThielemans:
let's look at this after #371

PR #386, opened 2020-05-01
pass CMAKE_BUILD_TYPE

rijobro:
fixes https://github.com/SyneRBI/SIRF-SuperBuild/issues/301.

Tested with:
```
rm -rf builds/STIR
cmake -DCMAKE_BUILD_TYPE=dfgdfg
make -j4 SIRF
```
Error (as expected):
```
loading initial cache file /Users/rich/Documents/Code/SIRF-SuperBuild/Build/builds/STIR/tmp/STIR-cache-dfgdfg.cmake
-- ccache found, so we will use this.
Unknown build type "dfgdfg".
CMake Error at CMakeLists.txt:29 (message):
  Allowed values are Debug, Release, RelWithDebInfo (case-insensitive).
```


KrisThielemans:
looks ok. All CIL jobs fail because of #390.

However, I think we will want to have an option to change this per-project. Seems simple enough with the macros, but let's do that after the release though, so please create an issue for that.

Issue #383, opened 2020-04-29
BUILD_SIRF_Registration and BUILD_SIRF=OFF


KrisThielemans:
When setting  `BUILD_SIRF=OFF`, it will be re-enabled because `BUILD_SIRF_Registration=ON`.

KrisThielemans:
@rijobro if we could fix this one as well, we should be almost ready i hope. sorry to load this all into you

Issue #389, opened 2020-05-05
CIL_FRAMEWORK_TESTS_1 are failing with libcilacc.so


KrisThielemans:
All +cil jobs are failing in https://github.com/SyneRBI/SIRF-SuperBuild/pull/387. 
https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/builds/682882279 see eg.
https://api.travis-ci.org/v3/job/682882281/log.txt
```
5/8 Test #5: CIL_FRAMEWORK_TESTS_1 ............***Failed    0.24 sec
EE
======================================================================
ERROR: test_DataContainer (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_DataContainer
Traceback (most recent call last):
  File "/usr/lib/python3.5/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/usr/lib/python3.5/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/build/SyneRBI/SIRF-SuperBuild/sources/CCPi-Framework/Wrappers/Python/test/test_DataContainer.py", line 23, in <module>
    from ccpi.framework import DataContainer
  File "/home/travis/build/SyneRBI/SIRF-SuperBuild/INSTALL/python/ccpi/framework/__init__.py", line 28, in <module>
    from .framework import DataContainer
  File "/home/travis/build/SyneRBI/SIRF-SuperBuild/INSTALL/python/ccpi/framework/framework.py", line 47, in <module>
    cilacc = ctypes.cdll.LoadLibrary(dll)
  File "/usr/lib/python3.5/ctypes/__init__.py", line 425, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python3.5/ctypes/__init__.py", line 347, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libcilacc.so: cannot open shared object file: No such file or directory
```
Most likely we're still using CIL `master`  tag.

@paskino can this be sorted out at high priority please?

KrisThielemans:
no longer using `master`
https://github.com/SyneRBI/SIRF-SuperBuild/blob/dfd7e9795fe94457e86498a5fe7f9bb0f30c6d33/version_config.cmake#L157-L165
so I don't know what's happening

KrisThielemans:
[...]

PR #393, opened 2020-05-05
correct logic around building sirf and registration

rijobro:
fixes https://github.com/SyneRBI/SIRF-SuperBuild/issues/383


rijobro:
I can't really see how `BUILD_SIRF` is affected by `BUILD_SIRF_Registration`, it only resets the value of `BUILD_NIFTYREG `, and that doesn't affect `BUILD_SIRF` either. Anyway hopefully this PR will solve it.

KrisThielemans:
ok, probably what I saw was that I disabled NiftyReg and SIRF, and that it switched NiftyReg back on again. sorry about misleading you!

KrisThielemans:
usual failures because of #389. OSX ones are still in the queue but as there's nothing OS specific here, I'll merge.

Issue #388, opened 2020-05-04
git doesn't fetch new tags


KrisThielemans:
On 2.1 VM:
```
fatal: ambiguous argument 'rel_4.0.1': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
From https://github.com/UCL/STIR
   8942f4e6..2a799b93  Siemens_mCT_additions -> origin/Siemens_mCT_additions
   fd8b7555..dcfda6b1  master                -> origin/master
 * [new branch]        release_4             -> origin/release_4
 * [new branch]        release_4.0           -> origin/release_4.0
 * [new tag]           rel_4.0.0             -> rel_4.0.0
 * [new tag]           stir_rel_4.0.0        -> stir_rel_4.0.0
error: pathspec 'rel_4.0.1' did not match any file(s) known to git.
CMake Error at /home/sirfuser/devel/buildVM/builds/STIR/tmp/STIR-gitupdate.cmake:147 (message):
  Failed to checkout tag: 'rel_4.0.1'
```
_Originally posted by @rijobro in https://github.com/SyneRBI/SyneRBI_VM/issues/160#issuecomment-623390728_

KrisThielemans:
same problem when using  `origin/rel_4.0.1` in `version_config.cmake`.

KrisThielemans:
@rijobro wrote
> I ran git fetch --tags earlier (didn't think to create a snapshot, sorry) and then compilation was fine and all ctests passed.

I suppose you ran this in `sources/STIR`?

According to the [CMake doc](https://cmake.org/cmake/help/latest/module/ExternalProject.html) I think this should work. However, they do recommend using the hash.

@rijobro or @johannesmayer could you try to set `STIR_TAG` in `version_config.cmake` to a hash that it doesn't have yet, and see if that works (I guess you'll have to start from a new VM if you didn't snapshot)

rijobro:
> I suppose you ran this in sources/STIR?

By hash, you mean the hash of the commit? I'm fairly sure that would work, probably so would the branch name (`release_4`). I'll confirm tomorrow.

KrisThielemans:
Branch name is dangerous according to their doc, although `origin/release_4` might work. However, that branch is moving, so that's no solution.

let's try the hash. of course, then we'd have to do the same for the others.

weird though

KrisThielemans:
[...]

PR #394, opened 2020-05-06
Update stir branch

rijobro:
@KrisThielemans Do you want to switch the STIR tag to using the hash? Feel free to close this if you're happy with it as is.


KrisThielemans:
If it resolves #388 it's fine, but it looked like it didn't. (at least yesterday). in that case, I'd prefer the tag obviously (clearer).

if we merge it, I'd prefer to have a comment that this is is the hash for 4.0.1

KrisThielemans:
Closing as this didn't resolve the problem.

PR #396, opened 2020-05-06
check if default swig executable exists

### LINKED ISSUE #395: SWIG_EXECUTABLE should change with USE_SYSTEM_SWIG
With `USE_SYSTEM_SWIG=OFF` (default), `SWIG_EXECUTABLE` is set to:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/95c7041db663d3afb60f5501a40d288fc25da50f/SuperBuild/External_SWIG.cmake#L81

If you then switch to `USE_SYSTEM_SWIG=ON`, the value of `SWIG_EXECUTABLE` doesn't change, even if `${SWIG_INSTALL_DIR}/bin/swig` doesn't exist.

Perhaps this could be fixed by adding to the following such that we check if the value is the default and if the file doesn't exist, reset the cached value:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/95c7041db663d3afb60f5501a40d288fc25da50f/SuperBuild/External_SWIG.cmake#L85-L89

rijobro:
fixes #395.

### DIFF
--- a/SuperBuild/External_SWIG.cmake
+++ b/SuperBuild/External_SWIG.cmake
@@ -35,13 +35,20 @@ set(externalProjName ${proj})
 
 SetCanonicalDirectoryNames(${proj})
 
+if (WIN32)
+  set(SWIG_INSTALL_DIR ${SUPERBUILD_INSTALL_DIR}/SWIG-3.0.12)
+  set(SB_SWIG_EXECUTABLE ${SWIG_INSTALL_DIR}/swig.exe)
+else(WIN32)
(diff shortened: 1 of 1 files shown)


rijobro:
Tested this by switching the value of `USE_SYSTEM_SWIG` to `OFF` and `ON` and `SWIG_EXECUTABLE` switches between `/Users/rich/Documents/Code/SIRF-SuperBuild/Install/bin/swig` and `/usr/local/bin/swig`, respectively, as expected.

rijobro:
Non CIL builds pass, merging.

PR #400, opened 2020-05-22
updated to STIR 4.0.2

KrisThielemans:


casperdcl:
docker jobs are expected to run for local branches [travis-ci/SIRF-SuperBuild#1859](https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/builds/690003791). They won't run for PRs [travis-ci/SIRF-SuperBuild#1860](https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/builds/690003846).

In the case of an external PR there would be no local branch build.

If we don't want PRs from local branches to run two builds, that would be a different issue.

KrisThielemans:
thanks @casperdcl . makes sense. We can leave that for now then.

PR #401, opened 2020-05-26
clean-up Boost flags

KrisThielemans:
Introduce a variable Boost_CMAKE_ARGS, set in External_Boost.make, and use that everywhere. Main practical difference is that this will pass through any setting of `BOOST_ROOT` when `USE_SYSTEM_BOOST=ON`

I've amalgamated lines from elsewhere so this should be fine. Nevertheless, might be best to postpone a merge till after 2.2, unless a few of you can try this out. (not urgent)


SuperBuild/External_Boost.cmake
```diff
@@ -77,10 +77,31 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
...
     if(${USE_SYSTEM_${externalProjName}})
       find_package(${proj} ${${externalProjName}_REQUIRED_VERSION} REQUIRED)
-      message(STATUS "USING the system ${externalProjName}, using BOOST_LIBRARY_DIR=${BOOST_LIBRARY_DIR}")
+      set(Boost_CMAKE_ARGS -DBOOST_INCLUDEDIR:PATH=${Boost_INCLUDE_DIR})
+      if (Boost_LIBRARY_DIR_RELEASE)
+        # TODO might need to use Boost_LIBRARY_DIR_Debug for Debug builds
+        set(Boost_CMAKE_ARGS ${Boost_CMAKE_ARGS} -DBOOST_LIBRARYDIR:PATH=${Boost_LIBRARY_DIR_RELEASE})
+      endif()
+      if (BOOST_ROOT)
+        set(Boost_CMAKE_ARGS ${Boost_CMAKE_ARGS} -DBOOST_ROOT:PATH=${BOOST_ROOT})
```
rijobro, line 95: Here, I think you're saying, "if the user initially gave us `BOOST_ROOT`, then pass that onto the dependencies". Is there a variable generated by the `find_package(Boost)` that we could pass on?

rijobro: I say that because presumably the version of boost given by, say, `Boost_LIBRARY_DIR_RELEASE` could differ from `BOOST_ROOT`. Not that this is a new problem...

KrisThielemans: findBoost.cmake isn't in fantastic shape. It sets some cached variables and uses some others as input. We try to cover those in the lines above. They aren't sufficient generally (not on Windows anyway).

The rationale here is that if the user gave use `BOOST_ROOT`, presumably she'll want us to pass it through. Of course, a sophisticated user could afterwards adapt any boost cached variables and hope that those would be passed as well, which still wouldn't work with the current PR. However, if a user sets `Boost_LIBRARY_DIR_RELEASE` to something else than what was found from `BOOST_ROOT`, it seems asking for trouble anyway...

Possibly safer would be to figure out what cached variables are present after the `find_package(Boost)` and pass all of those on as well. Might be possible.

PR #402, opened 2020-05-26
Pass HDF5_ROOT through to projects

KrisThielemans:
... if it's defined and `USE_SYSTEM_HDF5=ON`


KrisThielemans:
This should be entirely harmless, but is quite useful.

PR #404, opened 2020-05-29
disable STIR HDF5 support by default

rijobro:
By default, disable STIR's HDF5 support. Avoids this current problem - https://github.com/SyneRBI/SIRF/issues/685.


rijobro:
If we wait for this PR to be green on travis, merge, then wait for SIRF master to be green, we could be waiting quite a while.

Should we wait or should I just merge this one directly? Seems harmless, but you never know... @KrisThielemans

KrisThielemans:
as this pr tests SIRF master already, there's no need for a double wait.

In any case, I'm happy to tag SIRF once the DEVEL jobs in this PR succeed

rijobro:
OSX builds seem to be hanging, but I'm happy as Linux are good and OSX was fine locally.

rijobro:
@evgueni-ovtchinnikov you can start tagging.

PR #407, opened 2020-06-04
disable hdf5 by default

rijobro:
In spite of https://github.com/SyneRBI/SIRF-SuperBuild/pull/404, STIR HDF5 support still being built on PRs, causing them to fail: https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/jobs/694165572#L22214-L22217.

I'm guessing this is because there are two bools controlling it in STIR. Set them to both use the same boolean SuperBuild value.


KrisThielemans:
There has never been a `DISABLE_HDF5_SUPPORT` as far as I can see. It certainly is `DISABLE_HDF5` in STIR `release_4` branch (it isn't yet anywhere else I believe).

KrisThielemans:
please remove the `*SUPPORT` variable therefore

PR #411, opened 2020-06-09
remove unsused fftw3 files

rijobro:


KrisThielemans:
I have no time to look at this now but removing this makes no sense to me. Gadgetron needs FFTW and I believe our MR projector as well.

rijobro:
Just removing empty files, the main FFTW files stay.

Issue #414, opened 2020-06-13
When building on 18.04 STIR is not built with ITK


ALEXJAZZ008008:
If I select for STIR to be built with ITK I get the following error when trying to use stir_math with ITK:

~~~~
FactoryRegistry: key ITK not found in current registry
2 possible values are:
Interfile
None

Using value corresponding to key "None"

WARNING: KeyParser warning: unrecognized keyword: itk output file format parameters


WARNING: KeyParser warning: unrecognized keyword: default extension


WARNING: KeyParser warning: unrecognized keyword: end itk output file format parameters

Error parsing output file format from /home/alex/Documents/jrmomo/interfile_to_pet/data/par/stir_math_ITK_output_file_format.par
~~~~

With this par file:

~~~~
output file format parameters:=
; sample .par file that can be used with stir_math to specify the
; output file format
;
; See the stir::OutputFileFormat hierarchy for possible values.
; here we use ITK (so this will only work if you linked STIR with ITK)
;
; Example usage to convert an image (in any format that STIR can read) to 
; a file format that ITK can write:
;
; convert input.hv to MetaIO
;    stir_math --output-format stir_math_ITK_output_file_format.par output.mhd input.hv
; use the default extension to determine the file format (here set too Nifti)
;    stir_math --output-format stir_math_ITK_output_file_format.par output input.hv

output file format type := ITK
ITK Output File Format Parameters:=
  ; use Nifti as output (could also use .mhdr, .nhdr etc)
  default extension:=.nii
End ITK Output File Format Parameters:=
END:=
~~~~

There are no build errors.

KrisThielemans:
it didn't find ITK when building. Please give more info. Also have a look at ITK_DIR in builds/STIR/build/CMakeCache.txt

ALEXJAZZ008008:
ITK_DIR does not exist in builds/STIR/build/CMakeCache.txt although DISABLE_ITK does and is set to ON which would explain my problem. In the SuperBuild there is no option to set DISABLE_ITK.

While looking I also found:

~~~~
//No help, variable specified on the command line.
DISABLE_HDF5:BOOL=ON

//disable use of HDF5 libraries
DISABLE_HDF5_SUPPORT:BOOL=OFF
~~~~

Which either seems wrong or redundant.

rijobro:
From SuperBuild `USE_ITK=ON`.

[...]

PR #415, opened 2020-06-16
Stir disable itk fix

ALEXJAZZ008008:
I believe the second clause of the enclosing if statement could be written better as an if else rather than setting the same thing twice but I wanted to stay as close to the code snippet in issue #414 as possible


rijobro:
Thanks. Have you tested this works? By first building with ITK `OFF`, and then switching to `ON`?

PR #420, opened 2020-07-04
changes to ITK building

### LINKED ISSUE #210: build less of ITK
In the spirit of #12, and because of time-outs on Travis https://github.com/CCPPETMR/SIRF/issues/286, we should try to reduce what modules are built from ITK, which [can be done via groups](https://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch2.html). I think STIR only needs `-DITKGroup_IO=On`.

Currently, we're building with `-DModule_ITKReview=ON`, which then drags in pretty much everything. This is currently [required for pet-rd-tools[(https://github.com/UCL/pet-rd-tools/issues/35). Until that is fixed, we could presumably only include that module if we're building pet-rd-tools, but that gets a bit messy.

KrisThielemans:
- add new option `ITK_MINIMAL_LIBS` (default to `ON`) to build less of ITK
    - allow use of option `ITK_EXTRA_CMAKE_ARGS`

    fixes #210


KrisThielemans:
One unrelated Travis failure on OSX (not in a job that uses ITK anyway), so merging

Issue #129, opened 2018-06-29
HDF5 conflicts


KrisThielemans:
we're getting conflicts between HDF5 versions. Easiest route might be to use static linking  when building SIRF and ISMRMRD.

KrisThielemans:
confirmed with @johannesmayer that HDF5 conflicts easily happen even on the VM. We had the system HDF5, the SuperBuild HDF5 and ITK comes with HDF5 as well. Matlab does as well. All different versions. And the GE

HDF5 library has a habit of causing segfaults when there's a version mismatch anywhere, so this needs to be avoided.

I lean now towards relying on a system HDF5 by default (i.e. to not make the situation worse), and introduce static linking. The latter should be doable by adding `-DHDF5_USE_STATIC_LIBRARIES=ON` in the `External*cmake`. I don't think I'd put it in SIRF's CMakeLists.txt.

Anyone any other opinions? @bathomas, @casperdcl ?

paskino:
Can't we rely only on the one we build?

KrisThielemans:
we're trying to, but failing :-; I did notice there were a few `HDF5_LIBRARIES` missing in the `External*` files. @johannesmayer, maybe you could generate a PR with those fixes to the SuperBuild. However, even then we had conflicts until we deleted all our own HDF5 files, and just relied on the system one. I now think that relying on the system one is best, unless we cannot (e.g. because it's too old).

KrisThielemans:
check if ITK can be build with a specific HDF5.

change option `USE_SYSTEM_HDF5` depending on `find_package`

KrisThielemans:
> check if ITK can be build with a specific HDF5.
yes: `-DITK_USE_SYSTEM_HDF5=ON`

KrisThielemans:
The situation on specifying HDF5 location is actually complicated.

At present we set `HDF5_ROOT`, `HDF5_INCLUDE_DIRS` and most of the time `HDF5_LIBRARIES`.  However, if `USE_SYSTEM_HDF5=OFF`, `HDF5_LIBRARIES` is actually empty.

[...]

PR #422, opened 2020-07-04
Update External_STIR.cmake for HDF5

### LINKED ISSUE #421: STIR build can pick up the wrong HDF5
Currently, HDF5 support in STIR is by default disabled. However, #419 tries to enable it, but found that STIR was build with a different (more recent) version of HDF5. See https://github.com/SyneRBI/SIRF-SuperBuild/pull/419#issuecomment-653741259

Cause is a missing `${HDF5_CMAKE_ARGS}` in `ExternalSTIR.cmake`

KrisThielemans:
- add `${HDF5_CMAKE_ARGS}`
- move `${Boost_CMAKE_ARGS}` to the same place for consistency

Fixes #421


KrisThielemans:
possibly (but unlikely) cause for #129

PR #427, opened 2020-07-26
Set HDF5_C_LIBRARIES when using USE_SYSTEM_HDF5=OFF

KrisThielemans:
This variable is used by ISMRMRD. If we don't set it, it seems that it can find an unrelated HDF5 version, causing problems.


KrisThielemans:
It's not clear when this variable should be set the way it is, but it doesn't seem to do any harm. Travis had unrelated problems, so merging

Issue #161, opened 2018-12-07
building gadgetron on VM with boost 1.68 causes undefined reference to `boost::system::detail::generic_category_instance'


evgueni-ovtchinnikov:
Building Gadgetron needs gcc6 and boost 1.65.0 We tried with 1.68.0, but then got
```
undefined reference to `boost::system::detail::generic_category_instance'
```
This might be related to https://github.com/boostorg/system/issues/26 as g++-6 defaults to C++-14 and therefore resolved by making sure that boost is built with gcc-6. This is actually #14.

KrisThielemans:
@evgueni-ovtchinnikov please try again with the new versiono fthe Superbuild. It works for me.

evgueni-ovtchinnikov:
`update_VM.sh` runs gcc-5.4.0

linking executable test_ismrmrd failed with error messages

    undefined reference to `boost::test_tools::tt_detail::report_assertion(boost::test_tools::assertion_result const&, boost::unit_test::lazy_ostream const&, boost::unit_test::basic_cstring<char const>, unsigned long, boost::test_tools::tt_detail::tool_level, boost::test_tools::tt_detail::check_type, unsigned long, ...)'
    undefined reference to `boost::unit_test::ut_detail::auto_test_unit_registrar::auto_test_unit_registrar(boost::unit_test::test_case*, boost::unit_test::decorator::collector&, unsigned long)'

had to use the same trick of setting (after failed update) `USE_SYSTEM_ISMRMRD=ON` to build

KrisThielemans:
Confirmed that this also happens with gcc 5.4 and older Gadgetron, e.g. https://github.com/gadgetron/gadgetron/commit/e7eb430673eb3272e8a821b51750c0a2a96dafed

It disappears when editing gadgetron's [CMakeLists.txt](https://github.com/gadgetron/gadgetron/blob/e7eb430673eb3272e8a821b51750c0a2a96dafed/CMakeLists.txt#L145) to force use `std=c++11` and not 14 (although this gives some warnings)

It seems therefore that we cannot use a packaged system boost 1.68, and that we'd have to build Boost ourselves, making sure that c++-14 is used.

sigh.

KrisThielemans:
Lengthy discussion at https://groups.google.com/forum/#!topic/boost-developers-archive/EWG5NVOZo_g Conclusion is that you need to build boost (or any C++ library) with the same C++ standard version as you own code.

[...]

Issue #428, opened 2020-07-26
Travis failure on OSX with Python 3.6


KrisThielemans:
It seems we have only 1 job running with Python 3 on OSX, but it is failing now. Example https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/jobs/711879526

Error occurs [here](https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/jobs/711879526#L4404-L4407)
```
$ $PY_EXE -m pip install --user -U nose codecov coveralls
...
Collecting docopt>=0.6.1
  Downloading docopt-0.6.2.tar.gz (25 kB)
    ERROR: Error [Errno 2] No such file or directory: '/usr/local/opt/python/bin/python3.7': '/usr/local/opt/python/bin/python3.7' while executing command python setup.py egg_info
ERROR: Could not install packages due to an EnvironmentError: [Errno 2] No such file or directory: '/usr/local/opt/python/bin/python3.7': '/usr/local/opt/python/bin/python3.7'
```

I have no clue why it attempts to find a python version on `/usr/local/opt` as that's not the python we're executing:
```
$ echo "Using Python executable $PY_EXE"
Using Python executable /usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/bin/python3.7
```

PR #429, opened 2020-07-26
set preprocessor define BOOST_ERROR_CODE_HEADER_ONLY

KrisThielemans:
We get linking problems with boost system if the C++ versions don't match (i.e. the one used to compile boost, and the current one), specifically with boost::system::detail::generic_category_instance

Setting this variable forces Boost to use an inline variable, as opposed to the instantiated variable in the library

Fixes https://github.com/SyneRBI/SIRF-SuperBuild/issues/161


KrisThielemans:
One unrelated Travis failure (due to #428), so merging.

@evgueni-ovtchinnikov this should fix #161.

PR #431, opened 2020-07-28
force Armadillo libraries to be installed in lib

### LINKED ISSUE #430: Armadillo can install into lib64 or elsewhere
Armadillo has in its `CMakeFiles.txt`
```cmake
# Adhere to GNU filesystem layout conventions
include(GNUInstallDirs)
```
This will follow [GNU conventions](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html). On our CentOS7 system, it means it will install in `lib64`, but all other packages are in `lib`, and our `env_*sh` scripts expect it to be there.

I guess we have to force it to install in `lib` anyway

KrisThielemans:
This override the GNUinstallation conventions, as none of the other packages uses this.

Fixes #430


KrisThielemans:
unrelated Travis failure due to #428, so merging

PR #432, opened 2020-07-28
[TRAVIS] change setting of PY_EXE on OSX, and use only python3

KrisThielemans:
Use the simpler mechanism introduced in SIRF earlier and which seems to work fine there.

Take the opportunity to sync parts of the comments with the SIRF .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -135,9 +135,9 @@ jobs:
  # osx
  - os: osx
    osx_image: xcode11.2
-   python: 2.7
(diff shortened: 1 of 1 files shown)


KrisThielemans:
hopefully fixes #428

KrisThielemans:
This fixed the Python3 case, but made all Python2 cases fallover due to mixed use of Python and libraries. This could be fun to sort out, but we are going to drop Python2 in any case, so I'll just check what happens when using Python3 on OSX all the time (as already done on SIRF travis)

PR #435, opened 2020-08-10
Remove RenameVariable macro

### LINKED ISSUE #433: STIR does not build SWIG if required to build Python wrapper
Currently, the dependency on SWIG is [checked ](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/SuperBuild/External_STIR.cmake#L68)using the old variable `BUILD_STIR_SWIG_PYTHON` rather than the renamed variable `STIR_BUILD_SWIG_PYTHON`.

paskino:
closes #433


paskino:
This will require to update all the `SuperBuild\External_` to use `RenameVariable` with the new behaviour.

KrisThielemans:
@paskino this PR looks perfectly alright to me. Of course, it doesn't do what the title says, but it does fix #433. Should we just merge this?

evgueni-ovtchinnikov:
should we wait until this PR is merged or can I tag and release SIRF-SuperBuild right away?

KrisThielemans:
I think this isn't high priority. If @paskino isn't sure, I'd be happy to postpone it till after tagging.

However, I'm assuming we're still testing docker. I'll write an email.

PR #449, opened 2020-12-22
Enable HDF5 for STIR by default

### LINKED ISSUE #447: STIR needs to (optionally) depend on HDF5
To be able to use GE files, STIR needs to be built with HDF5 support. We currently have a variable
https://github.com/SyneRBI/SIRF-SuperBuild/blob/0365a51041f885e652d560ad7d4848555ad19fe0/SuperBuild/External_STIR.cmake#L53
But the project doesn't depend on HDF5, meaning that  `HDF5_CMAKE_ARGS` won't necessarily be set (nor the libraries built of course)

KrisThielemans:
- set default for STIR_ENABLE_HDF5=On
- make STIR depend on HDF5 (if STIR_ENABLE_HDF5)
- build C++ version of HDF5 libraries as this is what STIR needs

fixes #447


KrisThielemans:
Travis failures on OSX due to unrelated #450 , so merging.

PR #466, opened 2021-01-06
fix docker pull rate limit

casperdcl:
- fixes https://github.com/SyneRBI/SyneRBI-admin/issues/1


KrisThielemans (changes requested):
looks fine. let's see what it does.

I suggest to move the comment `# for counting clones, excluding ours` down  1 or 2 lines

casperdcl:
> I suggest to move the comment `# for counting clones, excluding ours` down 1 or 2 lines

good spot

PR #473, opened 2021-01-15
fix/simplify Python CMake arguments

### LINKED ISSUE #472: PYTHON_LIBRARY is not properly passed to SIRF, due to SuperBuild CMake bug
Read https://github.com/SyneRBI/SIRF-SuperBuild/issues/467#issuecomment-760933231 and subsequent comments for clarification and explanation.

KrisThielemans:
- Created a variable `PYTHONLIBS_CMAKE_ARGS` that contains all settings to be used.

- No longer pass `PYTHON_INCLUDE_DIR` from `...DIRS`, nor` PYTHON_LIBRARY` from` ..LIBRARIES`, but from the variable itself. This caused problems when there are multiple files, and in particular debug and release Python libraries present.

Fixes #472


SuperBuild.cmake
```diff
@@ -232,8 +247,11 @@ option(BUILD_pet_rd_tools "Build pet_rd_tools" OFF)
 option(BUILD_CIL "Build CCPi CIL Modules and ASTRA engine" OFF)
 option(BUILD_CIL_LITE "Build CCPi CIL Modules" OFF)
 option(BUILD_NIFTYREG "Build NIFTYREG" ON)
-option(BUILD_SIRF_Contribs "Build SIRF-Contribs" ON)
-
+if (BUILD_PYTHON)
```
rijobro, line 250: I guess there's no guarantee that the contributions will always be limited to python.

rijobro (approved):
haven't tested it obviously, but looks good to me!

SuperBuild.cmake
KrisThielemans: that's true, but at the moment it does https://github.com/SyneRBI/SIRF-SuperBuild/blob/199921c3305a1acd1c95d8de4ebfdcdd00be286b/SuperBuild/External_SIRF-Contribs.cmake#L48
which fails  if `PYTHON_DOEST` isn't set.

Maybe a better fix would be to fix that line

KrisThielemans:
@rijobro  I have now used a different strategy for `SIRF-contribs`

AnderBiguri:
Sorry I didn't test this before, but I get:

```
CMake Error at SuperBuild.cmake:146 (if):
  if given arguments:

    "EXISTS" "optimized" "C:/Users/Ander/AppData/Local/Programs/Python/Python38/libs/python38.lib" "debug" "C:/Users/Ander/AppData/Local/Programs/Python/Python38/libs/python38_d.lib"

  Unknown arguments specified
  CMakeLists.txt:37 (include)
```

KrisThielemans:
bugger. I guess I was too optimistic. I believe this is due to yet another bug in `FindHDF5.cmake` (it shouldn't set `PYTHON_LIBRARY` like this, but apparently does). Could you try by putting some quotes around it?
```
if (EXISTS "${PYTHON_LIBRARY}")
```
also for the include files.

AnderBiguri:
Yup, that was the issue:
https://stackoverflow.com/questions/39707772/cmake-error-if-given-arguments-followed-by-parenthesis-not-equals-and-si/39733128

Issue #418, opened 2020-06-29
env_ccppetmr.csh/.sh -> env_ccpsynerbi.csh/.sh


rijobro:
We still have `env_ccppetmr.csh.in/.sh.in` and still install `env_ccppetmr.csh/.sh`. Should be `env_ccpsynerbi.csh/.sh`.

KrisThielemans:
indeed. we discussed this somewhere, but I cannot find it.

The suggestion was to rename as you suggest, but to either copy or [symlink](https://cmake.org/cmake/help/v3.15/command/file.html#create-link) to the old name at least for the next version.

PR #481, opened 2021-01-25
Update cil 2011

paskino:
Changes to build CIL 20.11, which is backward incompatible.

All CIL unittests will be added, once [this](https://github.com/vais-ral/CCPi-Framework/issues/745) issue is fixed. Also I noticed a problem with [this](https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/test/test_tiff_readerwriter.py#L22) line with CTest, probably it's better to create a temporary directory.


paskino:
Absolutely, as a matter of facts I create both `env_ccpsynerbi.sh` and `env_ccppetmr.sh`, they'll coexist for a transition period...

But we might as well make it a hard change without transition period.

Reference to any events really happened is purely casual.

KrisThielemans:
In all honesty, I find a PR with the title "update cil 2011" that messes with `ccp_env*.sh` quite confusing... I see no reason why this should be the case. Having this discussion here will just make things harder to find back.

KrisThielemans:
> as a matter of facts I create both env_ccpsynerbi.sh and env_ccppetmr.sh, they'll coexist for a transition period..

#418 suggests a link. I'd prefer that, as it makes it clearer what's happening. Of course, it might fail if there's an existing file there already.

> But we might as well make it a hard change without transition period.

...and get any user (do we have any?) annoyed.

paskino:
py27 builds fail.

KrisThielemans:
yes. We have to drop Python 2 support and testing therefore. Also xenial. We will need to announce this on the mailing list, put it in the various `CHANGES`, UsersGuides, edit the `LongTermPlan` etc. Can you create an issue for these things in both SIRF and here?

Trying to sort Travis out for SIRF at https://github.com/SyneRBI/SIRF/pull/855. Once done, we'll know what to put here.

paskino:
@evgueni-ovtchinnikov @KrisThielemans, is it all right to bump the SIRF version that we build here so that unit tests will actually pass? v2.2.0 has a known bug.

KrisThielemans:
>  is it all right to bump the SIRF version that we build here so that unit tests will actually pass? v2.2.0 has a known bug.

no it is not. Discussion elsewhere. you need to set the default SIRF version to a recent hash in `version_config.cmake`.

paskino:
I exactly meant to choose a more recent SIRF commit hash than `v2.2.0` and save it in `version_config.cmake`. 
This new commit hash will go in the SuperBuild master.

So I was wondering if that'd be ok.

KrisThielemans:
yup!

paskino:
closes #489

paskino:
Except for the[ xenial build](https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/jobs/758342740), all other builds work fine. And the error is in the `apt` phase.

@KrisThielemans Can I merge?

KrisThielemans (approved):
go ahead and merge!

PR #483, opened 2021-01-25
creates env_ccppetmr.* symbolic links

### LINKED ISSUE #418: env_ccppetmr.csh/.sh -> env_ccpsynerbi.csh/.sh

paskino:
closes #418 

As discussed also in #481, we create the environment scripts with names `env_ccpsynerbi.sh` and `env_ccpsynerbi.csh` and  symbolic links `env_ccpsynerbi.sh` and `env_ccpsynerbi.csh` pointing to the former.


paskino:
This requires `CMake >3.14`

KrisThielemans (changes requested):
Thanks, I' feel that bumping CMake to 3.14 for the sake of a link is probably too harsh. Can you instead check the CMake version, and make a copy of it's too old? sorry.

This PR will need a line in CHANGE.md and updates to our installation instructions as well.

KrisThielemans (changes requested):
Looks great. However, I think I prefer to call them `env_synerbi.*sh` , which is the name of the org. (Just looking ahead for when we're no longer a CCP...)

paskino:
BTW why not `env_sirf.sh`? It would make more sense to me.

KrisThielemans:
> BTW why not `env_sirf.sh`? It would make more sense to me.

to be discussed tomorrow. The SIRF-SuperBuild can be used to build STIR only (for instance)... confusing? sure

PR #488, opened 2021-01-30
Patch Gadgetron include files for RPE encoding

paskino:
Applies a patch to the installed Gadgetron include file `hoNFFT.h` that creates a compiler error in @johannesmayer [RPE-encoding](https://github.com/johannesmayer/SIRF/tree/rpe-encoding) branch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # ChangeLog
 ## v#.#.#
+- Patch Gadgetron include file hoNDFFT.h to remove spurious ".."
 - Environment files with name env_sirf.sh (and csh) are created. Symbolic links or copies with the previous name env_ccppetmr.sh (and csh) depending on the version of CMake available are made.
 - Enabled HDF5 support for STIR by default (build C++ libraries for HDF5)
 - Fix some issues with finding Python [#472](https://github.com/SyneRBI/SIRF-SuperBuild/issues/472)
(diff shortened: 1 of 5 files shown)


paskino:
Travis fails on all CIL tests as here we build SIRF 2.2.0 which had a bug in some algebra methods.

The other failure is due to a download error. 

I would merge.

KrisThielemans (changes requested):
Looks fine overall.

Did you test it works ok with the old Gadgetron (I can't see why not though).

I'm not super-excited to use Python for patching, but it seems it might be the most portable way of doing things. However, are you sure that 'PYTHON_EXECUTABLE` will be set if we switch `PYTHON` off (which is necessary for building MATLAB).

Finally, please include some doc in the `Gadgetron_include-patch.py` what it's doing, and ideally also a line in `CHANGES.md`

paskino:
Of course I have used Python to patch because it'd be the most portable (in my mind of a Python dev).

The patch is applied to the version of Gadgetron [currently ](b6191eaaa72ccca6c6a5fe4c0fa3319694f512ab)built currenlty in the SuperBuild.

I'll add a test to check if `$PYTHON_EXECUTABLE` is available and if not look for it.

KrisThielemans:
great. don't forget the documentation! Then merge.

KrisThielemans:
could just as well fix the Codacy issue (trailing white space)
https://app.codacy.com/gh/SyneRBI/SIRF-SuperBuild/pullRequest?bid=22015540&prid=6876424&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&
before you press the button

PR #495, opened 2021-02-22
Added parallelproj

### LINKED ISSUE #492: Add parallelproj projector
Add https://github.com/gschramm/parallelproj to the SuperBuild

KrisThielemans:
Add https://github.com/gschramm/parallelproj 

fixes #492


KrisThielemans:
We should now remove the `fill(0)` before forward projection. See https://github.com/gschramm/parallelproj/issues/7#issuecomment-783308373

KrisThielemans:
@paskino observed that `parallelproj` libraries were installed on his system in `install/lib64` rather than `install/lib`.

That'll be because @gschramm uses [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html).
This says it implements the [GNU standards](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html) but that doesn't talk about lib64 or not.  The [Filesystem Hierarchy Standard (FHS) document](https://www.pathname.com/fhs/pub/fhs-2.3.pdf) contains a paragraph

> This is commonly used for 64-bit or 32-bit support on systems which support multiple binary formats, but
require libraries of the same name. In this case, /lib32 and /lib64 might be the library directories, and
/lib a symlink to one of them

It's amazing that we only see this now. I think the best way to handle this is to add it to our `LD_LIBRARY_PATH` in the `env*` scripts.

gschramm:
@KrisThielemans, @paskino : I find the lib vs lib64 behavior of the GNUInstallDirs honestly not too useful, so we should discuss whether I should stick to it, or if we should use a more static approach (always use "lib").

paskino:
I tried to make a symbolic link between the `lib` and `lib64` directories with no success. Only copying the content of `lib64` to `lib` worked.

KrisThielemans:
I don't have a strong opinion about this. I guess the CMake/GNU people know what they're doing. Presumably they follow some architecture/OS conventions. you need `lib64` and `lib` where building both 64 and 32bit versions of course. We have never thought about this for STIR/SIRF, and nobody ever complained. It's entirely conceivable that if we'd ever get STIR/SIRF into debian, they'd tell us to use `GNUInstallDirs` anyway.

Whatever you choose, the SIRF-Superbuild has to cope with `lib64`.

gschramm:
ok. let's keep GNUInstallDirs. @KrisThielemans are there also official ways to obtain the GNUInstallDirs on a given platform? That would be useful for my python interface where I prefer to install the libs not systemwide, but rather only a conda virt. env. But there the problem of "lib" vs "lib64" remains. And I would prefer to find a solution that avoids required env. variables. 

At the moment I am using "glob" to find the libs but this is for sure not the best practice.

KrisThielemans:
Not sure what you're asking. Obviously you have it in `CMAKE_INSTALL_LIBDIR`. If you want to have it outside of that, all bets are off I guess.

Installing libraries in appropriate locations for Python is beyond what I know. If you create a `conda` feedstock, I'm assuming that `conda build` takes care of these things. If you use `setup.py`, you'll have to follow their recommendations. Sorry that I cannot help more on these things.

KrisThielemans:
Once @https://github.com/gschramm/parallelproj/pull/8 is merged (@gschramm let us know), this should be merged as well. I've checked that it works with CUDA well.

paskino (approved):

Issue #501, opened 2021-02-25
xenial builds seem to always fail


paskino:
shall we disable xenial travis build?

KrisThielemans:
sure, more to be done, as mentioned in #485 and the pointers there. But can start with disabling xenial.

KrisThielemans:
closed with 9fa3ce9

PR #503, opened 2021-03-16
Add Gadgetron dependency

johannesmayer:
Add Gadgetron as a dependency if BUILD_Gadgetron=ON

- Gadgetron is added as dependency to SIRF if it is build.
- ctest on Travis are run with  `--output-on-failure`

### DIFF
--- a/SuperBuild/External_CCPi-Framework.cmake
+++ b/SuperBuild/External_CCPi-Framework.cmake
@@ -60,7 +60,7 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
           -DPYTHON_DEST_DIR:PATH=${PYTHON_DEST}
           -DOPENMP_INCLUDES:PATH=${OPENMP_INCLUDES}
           -DCIL_VERSION:STRING=${${proj}_TAG}
-      INSTALL_COMMAND ${CMAKE_COMMAND} --build . --target install && ${CMAKE_COMMAND} -E copy_directory ${${proj}_SOURCE_DIR}/Wrappers/Python/data ${SUPERBUILD_INSTALL_DIR}/share/cil/ && ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/patches/cil-patch.py ${${proj}_SOURCE_DIR}/Wrappers/Python/cil/utilities/dataexample.py ${PYTHON_DEST}/cil/utilities/dataexample.py
+      INSTALL_COMMAND ${CMAKE_COMMAND} --build . --target install && ${CMAKE_COMMAND} -E copy_directory ${${proj}_SOURCE_DIR}/Wrappers/Python/data ${SUPERBUILD_INSTALL_DIR}/share/cil/ && ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/patches/cil-patch.py ${${proj}_SOURCE_DIR}/Wrappers/Python/ccpi/utilities/dataexample.py ${PYTHON_DEST}/ccpi/utilities/dataexample.py
       DEPENDS ${${proj}_DEPENDENCIES}
     )
 
--- a/SuperBuild/External_CCPi-Framework.cmake
+++ b/SuperBuild/External_CCPi-Framework.cmake
@@ -60,7 +60,7 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
           -DPYTHON_DEST_DIR:PATH=${PYTHON_DEST}
           -DOPENMP_INCLUDES:PATH=${OPENMP_INCLUDES}
           -DCIL_VERSION:STRING=${${proj}_TAG}
-      INSTALL_COMMAND ${CMAKE_COMMAND} --build . --target install && ${CMAKE_COMMAND} -E copy_directory ${${proj}_SOURCE_DIR}/Wrappers/Python/data ${SUPERBUILD_INSTALL_DIR}/share/cil/ && ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/patches/cil-patch.py ${${proj}_SOURCE_DIR}/Wrappers/Python/ccpi/utilities/dataexample.py ${PYTHON_DEST}/ccpi/utilities/dataexample.py
+      INSTALL_COMMAND ${CMAKE_COMMAND} --build . --target install && ${CMAKE_COMMAND} -E copy_directory ${${proj}_SOURCE_DIR}/Wrappers/Python/data ${SUPERBUILD_INSTALL_DIR}/share/cil/ && ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/patches/cil-patch.py ${${proj}_SOURCE_DIR}/Wrappers/Python/cil/utilities/dataexample.py ${PYTHON_DEST}/cil/utilities/dataexample.py
       DEPENDS ${${proj}_DEPENDENCIES}
     )
 
--- a/SuperBuild/External_SIRF.cmake
+++ b/SuperBuild/External_SIRF.cmake
@@ -25,7 +25,7 @@ set(proj SIRF)
 # Set dependency list
 set(${proj}_DEPENDENCIES "Boost;HDF5;ISMRMRD;FFTW3;SWIG")
 
-if (${BUILD_GADGETRON})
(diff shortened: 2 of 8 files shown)


SuperBuild/External_SIRF.cmake
```diff
@@ -24,6 +24,10 @@ set(proj SIRF)
 
 # Set dependency list
 set(${proj}_DEPENDENCIES "Boost;HDF5;ISMRMRD;FFTW3;SWIG")
+
+if (${BUILD_GADGETRON})
+    set(${proj}_DEPENDENCIES "${${prof}_DEPENEDCIES};Gadgetron")
```
KrisThielemans, line 29:
```suggestion
    set(${proj}_DEPENDENCIES "${${proj}_DEPENDENCIES};Gadgetron")
```

version_config.cmake
```diff
@@ -178,10 +178,12 @@ mark_as_advanced(DEVEL_BUILD)
 # with devel build it uses latest version of upstream packages
 # otherwise uses the versions for current SIRF
 
-set(DEFAULT_SIRF_URL https://github.com/SyneRBI/SIRF )
+#set(DEFAULT_SIRF_URL https://github.com/SyneRBI/SIRF )
+set(DEFAULT_SIRF_URL https://github.com/johannesmayer/SIRF.git )
```
KrisThielemans, line 182: can't have that later on of course

version_config.cmake
```diff
@@ -178,10 +178,12 @@ mark_as_advanced(DEVEL_BUILD)
...
 if (DEVEL_BUILD)
 
-  set (DEFAULT_SIRF_TAG origin/master)
+  #set (DEFAULT_SIRF_TAG origin/master)
+  set (DEFAULT_SIRF_TAG origin/rpe-encoding)
```
KrisThielemans, line 186: as above

SuperBuild/External_CCPi-Framework.cmake
```diff
@@ -60,7 +60,7 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
           -DPYTHON_DEST_DIR:PATH=${PYTHON_DEST}
           -DOPENMP_INCLUDES:PATH=${OPENMP_INCLUDES}
           -DCIL_VERSION:STRING=${${proj}_TAG}
-      INSTALL_COMMAND ${CMAKE_COMMAND} --build . --target install && ${CMAKE_COMMAND} -E copy_directory ${${proj}_SOURCE_DIR}/Wrappers/Python/data ${SUPERBUILD_INSTALL_DIR}/share/cil/ && ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/patches/cil-patch.py ${${proj}_SOURCE_DIR}/Wrappers/Python/cil/utilities/dataexample.py ${PYTHON_DEST}/cil/utilities/dataexample.py
+      INSTALL_COMMAND ${CMAKE_COMMAND} --build . --target install && ${CMAKE_COMMAND} -E copy_directory ${${proj}_SOURCE_DIR}/Wrappers/Python/data ${SUPERBUILD_INSTALL_DIR}/share/cil/ && ${PYTHON_EXECUTABLE} ${CMAKE_SOURCE_DIR}/patches/cil-patch.py ${${proj}_SOURCE_DIR}/Wrappers/Python/ccpi/utilities/dataexample.py ${PYTHON_DEST}/ccpi/utilities/dataexample.py
```
KrisThielemans, line 63: are you sure about this?

KrisThielemans (changes requested):

SuperBuild/External_SIRF.cmake
```diff
@@ -24,6 +24,10 @@ set(proj SIRF)
 
 # Set dependency list
 set(${proj}_DEPENDENCIES "Boost;HDF5;ISMRMRD;FFTW3;SWIG")
+
+if (${BUILD_GADGETRON})
+    set(${proj}_DEPENDENCIES "${${proj}_DEPENEDCIES};Gadgetron")
```
KrisThielemans, line 29: still wrong

SuperBuild/External_CCPi-Framework.cmake
johannesmayer: I don't understand what is correct. This location is different between the CCPi-Framework `master` branch and the tag `v20.09.1` which is always checked out on my VM. 

I reverted it, s.t. if CCPi-Framework `origin/master` is checked out it is correct.

SuperBuild/External_SIRF.cmake
johannesmayer: Git damnit, I'm sorry :D

SuperBuild/External_CCPi-Framework.cmake
KrisThielemans: @paskino could advice I guess

KrisThielemans:
The first job failed, but that's a xenial build that always fails #501 . However, the `DEVEL` build fails as well with https://travis-ci.com/github/SyneRBI/SIRF-SuperBuild/jobs/491517076#L25296
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.

Please set them or make sure they are set and tested correctly in the CMake files:
GT_CPUCORE
    linked by target "cgadgetron" in directory /home/travis/build/SyneRBI/SIRF-SuperBuild/sources/SIRF/src/xGadgetron/cGadgetron
GT_FFT
    linked by target "cgadgetron" in directory /home/travis/build/SyneRBI/SIRF-SuperBuild/sources/SIRF/src/xGadgetron/cGadgetron
GT_LOG
    linked by target "cgadgetron" in directory /home/travis/build/SyneRBI/SIRF-SuperBuild/sources/SIRF/src/xGadgetron/cGadgetron
GT_NFFT
    linked by target "cgadgetron" in directory /home/travis/build/SyneRBI/SIRF-SuperBuild/sources/SIRF/src/xGadgetron/cGadgetron
```
I guess that's a problem in your PR.

I suggest that you revert the lines that point to your PR and only keep the dependency update. Then we can merge this one, and you can keep tinkering away on the PR.

johannesmayer:
Thanks for the suggestion, I can do that.

SuperBuild/External_CCPi-Framework.cmake
paskino: Up to CIL 20.09 the python package lived under `ccpi`, from version 20.11 we have changed it to `cil`. So there is a backward incompatibility for CIL. 
In master the CIL version is [20.11.2](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/version_config.cmake#L166)

Finally I don't understand why the line had to be changed, but I suspect it is a local problem on @johannesmayer 's installation.

johannesmayer: yes that's correct. My SuperBuild checks out the old CIL tag and so it fails. I just don't change it here.

paskino:
I just found the reason of the failure: the dependency on Gadgetron was added if `BUILD_GADGETRON` flag was true, but the correct spelling is `BUILD_Gadgetron`. So it would always return false and never add Gadgetron in the pre-requisites.

paskino:
The build worked now, but the unit tests fail with this error
```
ImportError: /home/travis/build/SyneRBI/SIRF-SuperBuild/INSTALL/python/sirf/_pygadgetron.so: undefined symbol: cGT_getGRPETrajecotry
```

johannesmayer:
Ok thanks for the hint, I will sort it out... apparently the exclusion of non-cartesian code is still imperfect 👎

paskino:
but that error is coming from the `DEVEL_BUILD=ON`, that is when you build the RPE encoding, oder?

johannesmayer:
that is correct

KrisThielemans:
perfect! if you now revert the `version_config` changes, this should be good to merge in my opinion

johannesmayer:
I checked again and saw that the tests [unfortunately] took less than 10 minutes now and they succeeded. 
I hope that the line I added to avoid Travis timeout actually works when they take longer because it is the only line of "Travis code" that I ever wrote.

KrisThielemans:
> I hope that the line I added to avoid Travis timeout actually works when they take longer because it is the only line of "Travis code" that I ever wrote.

I don't really know, but
```
/3 Test #3: SIRF_TESTS .......................   Passed  636.00 sec
```
looks good :-)

Please add a line in `CHANGES` (with `[ci skip]').

ok to squash-merge this one?

johannesmayer:
Yes please, squash merge.

KrisThielemans (approved):

Issue #469, opened 2021-01-13
NiftyPET installation


KrisThielemans:
NiftyPET's installation procedure has improved to be CMake only, see https://github.com/NiftyPET/NIPET#external-cmake-projects.

We can remove [External_NiftyPET.cmake](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/CMake/install_NiftyPET.cmake) therefore and simplify [](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/SuperBuild/External_NiftyPET.cmake), but only after https://github.com/UCL/STIR/issues/795 is resolved.

KrisThielemans:
In https://github.com/UCL/STIR/issues/795#issuecomment-759493898 @casperdcl  said
> NIPET code is strongly coupled to the cpython wrapping API, so building it requires python. I think the current (hacky) way STIR and/or the SuperBuild goes about things is directly look for source code and (re)builds it itself. While some of the source code is python-binding-free, I'm not sure if it is really building it that way.

I haven't checked how @rijobro builds it (link above), but now I'm somewhat worried about Python conflicts. Will any Python do to build NiftyPET now, or does it have to be Python 3?

casperdcl:
`nipet<2.0.0` is py2-only, while `nipet>=2.0.0` is py3-only. The NIPET libraries which STIR is linking against https://github.com/UCL/STIR/blob/fe1af133aee5c44ce9fecd50993612f47d274b44/src/cmake/FindNiftyPET.cmake#L41-L45 are python modules so would've needed python to build.

KrisThielemans:
The python requirements do complicate things. As we are still supporting Python2, largely due to the VM not being updated, it seems that we have to do some `if/then/else` stuff sadly, both in STIR and here. oh well.

KrisThielemans:
In fact, on second thought, we're not building NiftyPET on the VM anyway (as no cuda), so maybe we just drop support for nipet<2.0.0. That is fine for me, as long as STIR still "works".

PR #504, opened 2021-03-18
switch default docker Python to miniconda3

KrisThielemans:
Use python3

### DIFF
--- a/docker/user_python-ubuntu.sh
+++ b/docker/user_python-ubuntu.sh
@@ -8,7 +8,7 @@ PYTHON="${2:-miniconda}"
 case "$PYTHON" in
 miniconda)
   # miniconda
-  curl https://repo.anaconda.com/miniconda/Miniconda2-latest-Linux-x86_64.sh > miniconda.sh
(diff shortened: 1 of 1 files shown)


KrisThielemans:
Only expected problem with NiftyPET I guess, due to #469

casperdcl (approved):
Sure. Regarding NiftyPET; should be easy to fix. Simply adding the cuda toolkit and then doing `pip install niftypet` would work fine, but as you say using the new `cmake` support (#469) may be required too for SIRF.

KrisThielemans:
@casperdcl due to (y)our changes to reduce Travis overload, the `docker` jobs actually don't run on a PR (only `docker-core`), meaning that this PR is effectively not tested, see e.g.
https://travis-ci.com/github/SyneRBI/SIRF-SuperBuild/builds/220508500
and so we won't see problems like #505.
 
We could change the `.travis` temporarily, or just merge to `master` and see. What do you think would be best? (without creating a lot of work!)

casperdcl:
I'd say mergig is fine - the docker images don't actually build any gpu stuff atm anyway (they only optionally comnect the gpu so users could install gpu code themselves).

PR #510, opened 2021-04-07
Bump cil tag

paskino:
updates to latest CIL release [21.0.0](https://github.com/vais-ral/CCPi-Framework/releases/tag/v21.0.0)

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -15,7 +15,7 @@ jobs:
         # the agent machine operating systems
         os: [ubuntu-latest, ubuntu-18.04]
         DEVEL_BUILD: ["OFF", "ON"]
-        EXTRA_BUILD_FLAGS: ["-DUSE_ITK=ON"]
+        EXTRA_BUILD_FLAGS: ["-DUSE_ITK=ON", "-DBUILD_CIL_LITE=ON"]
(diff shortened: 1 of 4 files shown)


KrisThielemans:
Please merge `master` in here such that we can see Actions in action 😄

paskino:
The failing CIL test needs investigating as it is running with numpy 1.20 and we've never tested it.

paskino:
@KrisThielemans I fixed the failure in https://github.com/vais-ral/CCPi-Framework/pull/843 

So if that is the only failure, we will merge this PR.

paskino:
Now that I disabled the `CIL_FRAMEWORK_TEST_6` this all works. However, current CIL master with the same unittest would [pass](https://github.com/SyneRBI/SIRF-SuperBuild/runs/2461523200?check_suite_focus=true)

I will reinstate the unittest after we've tagged CIL.

KrisThielemans (approved):

KrisThielemans:
any reason to not run all tests with `CIL_LITE=ON`? The number of tests is proliferating quickly of course.

On the other hand, need to test full CIL?

paskino:
Of course the requirements have changed in between this PR and now. I'll add `BUILD_CIL=ON` instead `BUILD_CIL_LITE=ON`

KrisThielemans (commented):
sure, let's go with this for now

PR #512, opened 2021-04-07
copy Gadgetron config file if not exists

### LINKED ISSUE #490: Gadgetron needs the user to create a config file
Once all the build has finished, users may want to start Gadgetron and they can't because of a missing config file.

Shall we create the appropriate config file in case it doesn't exist already? This would solve the problem for many.

paskino:
closes #490

### DIFF
--- a/SuperBuild/External_Gadgetron.cmake
+++ b/SuperBuild/External_Gadgetron.cmake
@@ -121,7 +121,9 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
       -DUSE_OPENMP:BOOL=${${proj}_ENABLE_OPENMP}
     DEPENDS
         ${${proj}_DEPENDENCIES}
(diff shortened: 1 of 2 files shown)


paskino:
I have tested on the VM and it works. I am working at a github action for CI.

KrisThielemans (commented):
ok, but why not simply
```cmake
install(FILES wherever/gadgetron.xml.example RENAME ${SUPERBUILD_INSTALL_DIR}/share/gadgetron/config/gadgetron.xml)
```
or whatever the syntax is?

paskino:
The desired behaviour is to create the Gadgetron config file if it does not already exist, do nothing otherwise.
This because a user might have modified the Gadgetron config file. 

I don't know how to achieve this with CMake.  The proposed change would have the desired behaviour and be _platform indipendent_ as it is relying on _just_ having Python on the target machine.

KrisThielemans:
>  platform indipendent as it is relying on just having Python on the target machine.

sounds like you've never developed on Windows 😄 . Some Windows users will have MATLAB but no Python... But of course, this is for Gadgetron and you are using Python for the patching already.

> The desired behaviour is to create the Gadgetron config file if it does not already exist, do nothing otherwise.
This because a user might have modified the Gadgetron config file. I  don't know how to achieve this with CMake. B

good point. CMake's `install` will compare the time-stamps and overwrite if the file to be installed is newer. Good enough? If not, go ahead and merge.

Using `INSTALL` will be better if we ever got to using CPack.

paskino:
And Gadgetron won't build on Windows either!

PR #513, opened 2021-04-07
Add GitHub action and use Python3 in docker

paskino:
Creates a GitHub for CI.

Although it only checks one configuration, it is better than none as we have now in the SuperBuild.


paskino:
The build was [successful](https://github.com/paskino/SIRF-SuperBuild/runs/2289222005?check_suite_focus=true) in my fork.

.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,48 @@
...
+          export CXX="g++-7";
+          source ~/devel/install/bin/activate;
+          cmake -S ${GITHUB_WORKSPACE} ${BUILD_FLAGS} ${EXTRA_BUILD_FLAGS} ;
+    - name: build
+      shell: bash
+      run: 
+          export INSTALL_DIR=~/install;
+          cd ${GITHUB_WORKSPACE}/build;
+          source ~/devel/install/bin/activate;
+          cmake --build . -j 2;
```
KrisThielemans, line 44: not sure what they recommend for parallel builds

docker/build_python-ubuntu.sh
```diff
@@ -2,6 +2,6 @@
 set -ev
```
KrisThielemans, line 2: are you sure about the permission change for this file? (I didn't check what the numbers mean)

docker/ctest_sirf.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash
+# source env 
+export INSTALL_DIR=~/install
+source ${INSTALL_DIR}/bin/env_sirf.sh
+
+# append the path of virtualenv
+export PATH=/home/runner/.local/bin:$PATH
```
KrisThielemans, line 7: using `/home/runner` probably makes it specific to Github actions. Could be useful for docker tests etc as well obviously. Possibly `${HOME}` works

docker/ctest_sirf.sh
```diff
@@ -0,0 +1,22 @@
...
+source ~/devel/install/bin/activate
```
KrisThielemans, line 8:
```suggestion
source ${INSTALL_DIR}/bin/activate 
```

docker/ctest_sirf.sh
```diff
@@ -0,0 +1,22 @@
+#! /bin/bash
+# source env 
+export INSTALL_DIR=~/install
+source ${INSTALL_DIR}/bin/env_sirf.sh
+
+# append the path of virtualenv
+export PATH=/home/runner/.local/bin:$PATH
+source ~/devel/install/bin/activate 
+
+${INSTALL_DIR}/bin/gadgetron >& gadgetron.log&
```
KrisThielemans, line 10:
```suggestion
gadgetron >& gadgetron.log&
```
as it's already in the path

docker/user_python-ubuntu.sh
```diff
@@ -34,7 +34,7 @@ esac
 
 # Python (runtime)
 if [ -f requirements.txt ]; then
```
KrisThielemans, line 36: permission change in this file as well

KrisThielemans (changes requested):
Looks great.

you'll have to merge `master` of course due to the conflict. (it might prevent running the action)

docker/user_python-ubuntu.sh
paskino: stupid windows!

docker/ctest_sirf.sh
paskino: My plan is to move this whole thing to docker container, but since I got it running, I though to add it to the SuperBuild as it is (with changes) so that we have a form of CI

KrisThielemans: for sure, although I wouldn't say we're only going to do CI via docker builds. At least, that doesn't look like a good idea to me (although not sure).

whatever path we follow, I certainly agree that we have to do this in stages of course. I just wondered if using `$HOME` wouldn't immediately generalise it, but feel free to postpone to later if you're not sure.

docker/user_python-ubuntu.sh
KrisThielemans: strange, I don't think my "git for windows" does that.

.github/workflows/c-cpp.yml
paskino: so far so good!

docker/ctest_sirf.sh
paskino: virtualenv is installed in [`~/devel/install`](https://github.com/paskino/SIRF-SuperBuild/blob/build-github-action/.github/workflows/c-cpp.yml#L24)

docker/build_python-ubuntu.sh
paskino: no, this is because I edited on windows and sometimes this happens...
fixed

docker/user_python-ubuntu.sh
paskino: I don't know, this happens to me. I'm using Visual Studio Code and git from cygwin, a poisonous mixture?

docker/ctest_sirf.sh
paskino: What I meant is that this action is using all the docker scripts but re-bundles them together. It'd be better to use them as they were meant to, although @casperdcl is using them with docker compose.

If I'm confused by docker, docker compose sets the bar even higher.

paskino:
Let's wait for the [action](https://github.com/paskino/SIRF-SuperBuild/actions/runs/728654439) to run and be green then we can merge?

docker/ctest_sirf.sh
```diff
@@ -0,0 +1,22 @@
...
+cd $GITHUB_WORKSPACE/build
+ctest --output-on-failure; test_fail=$?
+# echo "----------- Killing gadgetron server"
+# killall gadgetron
+if [[ $test_fail -ne 0 ]]; then
+ #echo "----------- Test output"
+ # cat builds/SIRF/build/Testing/Temporary/LastTest.log
+ echo "----------- Last 70 lines of gadgetron.log"
+ tail -n 70 gadgetron.log
+fi
```
KrisThielemans, line 22:
```suggestion
fi
exit $test_fail
```
Otherwise we don't get the correct test status

KrisThielemans (changes requested):

KrisThielemans:
What about trying `ccache` already? Seems simple enough
```
- name: ccache
  uses: hendrikmuhs/ccache-action@v1
...

- name: build
  run: |
    export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
     ...
```

KrisThielemans:
Please also edit `CHANGES.md`. Most important is probably to say that the Docker build moved to `python3`. (a bit surprising to have that in this PR, what about change the title of this PR to say "Add GitHub action and use Python3 in docker"?)

paskino:
> What about trying `ccache` already? Seems simple enough
> 
> ```
> - name: ccache
>   uses: hendrikmuhs/ccache-action@v1
> ...
> 
> - name: build
>   run: |
>     export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
>      ...
> ```

I need to know what `ccache` is ...

Let me know when to merge.

KrisThielemans:
`ccache` is "easy". it works behind the scenes to keep a cache of compilations, such that the next build could be 10 times faster. Highly recommended to install it on your system (and if it's in the path, CMake will find it).

but we can put those lines in a 2nd PR of course., so if you feel like (squash)merging now, go ahead.

docker/ctest_sirf.sh
casperdcl: `docker-compose` is a replacement for `make` really. Except it's specific to `docker` and it works on any OS.

PR #516, opened 2021-04-10
[GH ACTION] 2 Ubuntus and DEVEL_BUILD=ON/OFF

KrisThielemans:


paskino:
ah! I got you started @KrisThielemans !!!

KrisThielemans:
This is working fine. `DEVEL_BUILD` fails due to https://github.com/gschramm/parallelproj/issues/11, which then cancels all the other builds. I guess best to set [fail-fast](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to `false`.

Let's do that once the `parallelproj` bug is fixed.

PR #526, opened 2021-04-29
add cil build in SERVICE+GPU docker build

paskino:
Builds CIL for the SERVICE+GPU docker build


casperdcl:
Does CIL require a GPU?

paskino:
yes, at least the ASTRA engine we use.

casperdcl (approved):
changed to simply include CIL when building with GPU support

paskino:
Has travis run?

KrisThielemans:
> > Has travis run?
> 
> Clearly not configured to do so on external PRs :)

this is how @casperdcl set-up Travis for docker, in an attempt to minimise runs/pushes. Most PRs have nothing to do with docker, but it doesn't make sense for those who do. Not sure how to handle that.

KrisThielemans:
I can't say I like b7ca9d77f2df0c2dbd23c4cfbaad1e603c105860. It's more concise, but it certainly surprising that GPU suddenly has CIL as well.

I'd rather choose one of these 2:
- revert back to @paskino's strategy of having an explicit `*cil.yml` to add CIL
- always build with CIL (but this might fail without GPU?)

casperdcl:
Well just temporarily change the config in `.travis.yml` - or just merge and hope for the best since it seems minor.

casperdcl:
>     * revert back to @paskino's strategy of having an explicit `*cil.yml` to add CIL

That was a bit broken. He was still `dpush` `*cil.yml` to old tags rather than creating a new one.

>     * always build with CIL (but this might fail without GPU?)

idk. I'm going to unsubscribe from this issue since it's something you can decide on. `@mention` me again to summon me back :)

paskino:
Merging this PR now, but let us revert this when a GH to create such image is in place. Or maybe this will not be needed as we plan to move to GH.

PR #530, opened 2021-05-07
Update cil repos

paskino:
CIL repos have been transferred to the TomographicImaging organisation. This PR reflects these changes.


KrisThielemans (commented):
I haven't checked the detail here, but would it make sense to build `CCPi-Regularisation` with `CIL_LITE`? Otherwise we can't use any CIL regularisers, unless we're willing to install Astra.

KrisThielemans:
@paskino, I've killed the previous Travis job. Should I kill the current one as well? I guess we could let it go if it's the final update.

paskino:
> I haven't checked the detail here, but would it make sense to build `CCPi-Regularisation` with `CIL_LITE`? Otherwise we can't use any CIL regularisers, unless we're willing to install Astra.

Absolutely. I'll add that. Not building Travis.

PR #533, opened 2021-05-11
docker: not download data automatically

paskino:
Does not download the SIRF-Exercises by default on first start. User needs to specify it.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -105,9 +105,10 @@ COPY service.sh /usr/local/bin/
 ENV DEBIAN_FRONTEND ''
 CMD ["/usr/local/bin/service.sh"]
 
-FROM service as multi
-COPY service.multi.sh /usr/local/bin/
-
 COPY download_data.sh /usr/local/bin/
 ENV DEBIAN_FRONTEND ''
 CMD ["/usr/local/bin/download_data.sh"]
+
+FROM service as multi
+COPY service.multi.sh /usr/local/bin/
+
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -107,3 +107,7 @@ CMD ["/usr/local/bin/service.sh"]
 
 FROM service as multi
 COPY service.multi.sh /usr/local/bin/
+
+COPY download_data.sh /usr/local/bin/
+ENV DEBIAN_FRONTEND ''
+CMD ["/usr/local/bin/download_data.sh"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -75,8 +75,8 @@ RUN rm user_python-ubuntu.sh requirements.txt
 
 # SIRF
 COPY user_sirf-ubuntu.sh .
-COPY devel/.ccache/ /opt/ccache/
-RUN ccache -o cache_dir=/opt/ccache
+# COPY devel/.ccache/ /opt/ccache/
+# RUN ccache -o cache_dir=/opt/ccache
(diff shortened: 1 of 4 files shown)


KrisThielemans (commented):
obviously you will need to create the `download_data.sh` script (or did you?)

I would also move the `COPY` lines up such that this script is accessible in all docker images (now I think it's only in the `service-multi`).

KrisThielemans:
by the way, `[ci skip]` skips both GHA and Travis. `[actions skip]` only GHA, `[travis skip]` only Travis

docker/download_data.sh
```diff
@@ -0,0 +1,7 @@
+
+# download the SIRF-Exercises data
+pushd /devel
+[ -d SIRF-Exercises ] || cp -a $SIRF_PATH/../../../SIRF-Exercises .
+which unzip || sudo apt-get install -yqq unzip
```
KrisThielemans, line 5: although this presumably worked, I wonder if it wouldn't be safer to add unzip to our list of apt packages

docker/download_data.sh
```diff
@@ -0,0 +1,7 @@
+
```
KrisThielemans, line 1: this syntax looks suspiciously bash to me, so...
```suggestion
#! /bin/bash
```

docker/download_data.sh
```diff
@@ -0,0 +1,7 @@
...
+# download the SIRF-Exercises data
+pushd /devel
```
KrisThielemans, line 3: as it's in a script, this can now just be `cd /devel` and the `popd` removed at the end

docker/download_data.sh
```diff
@@ -0,0 +1,7 @@
...
+[ -d SIRF-Exercises ] || cp -a $SIRF_PATH/../../../SIRF-Exercises .
```
KrisThielemans, line 4: this stuff copies the SIRF-Exercises to /devel. This should not be in a `download_data.sh` script. I'm not 100% sure where it should be to be honest, but I'm fairly sure that the `service` image assumes that the `SIRF-exercises` are in `/devel`, so starting jupyter is now probably falling over as it cannot find things.

So, this line should either by kept in `service.sh` and `download_data.sh` only be copied in the `service` section, or (I think better but up for discussion) be moved to the `user_sirf-ubuntu.sh` and keep the `download_data.sh` in SIRF.

KrisThielemans (changes requested):

docker/Dockerfile
```diff
@@ -75,8 +75,8 @@ RUN rm user_python-ubuntu.sh requirements.txt
 
 # SIRF
 COPY user_sirf-ubuntu.sh .
-COPY devel/.ccache/ /opt/ccache/
-RUN ccache -o cache_dir=/opt/ccache
+# COPY devel/.ccache/ /opt/ccache/
+# RUN ccache -o cache_dir=/opt/ccache
```
KrisThielemans, line 79: re-enable this

docker/Dockerfile
```diff
@@ -107,3 +107,7 @@ CMD ["/usr/local/bin/service.sh"]
 
 FROM service as multi
 COPY service.multi.sh /usr/local/bin/
+
+COPY download_data.sh /usr/local/bin/
+ENV DEBIAN_FRONTEND ''
+CMD ["/usr/local/bin/download_data.sh"]
```
KrisThielemans, line 113: this needs to be moved up, as now it will only be available in the service-multi image (as far as I understand)

PR #541, opened 2021-05-20
set SIRF version 3.0.0

paskino:


version_config.cmake
```diff
@@ -206,7 +206,7 @@ if (DEVEL_BUILD)
 
 else()
   # just past 3.0.0-rc.1 due to Singles fix
```
KrisThielemans, line 208: could just as well delete this line with ` [ci skip]' once you know it works

paskino: absolutely!

PR #548, opened 2021-06-03
added environment variable CIL_VERSION required at install

paskino:
added environment variable CIL_VERSION required at install otherwise the build fails


paskino:
I'll be on leave tomorrow 4/6 so if you are happy please merge.

Issue #547, opened 2021-06-02
Docker: problems with server GPU


paskino:
I've run 
```

docker pull synerbi/sirf:service-gpu

./sirf-compose-server-gpu up -d sirf

Creating network "docker_default" with the default driver
Creating sirf ... error

ERROR: for sirf  Cannot create container for service sirf: Unknown runtime specified nvidia

ERROR: for sirf  Cannot create container for service sirf: Unknown runtime specified nvidia
ERROR: Encountered errors while bringing up the project.
```

My machine has 
```
[edo@vishighmem01 docker]$ nvidia-smi
Wed Jun  2 14:54:44 2021
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 418.52.18    Driver Version: 418.52.18    CUDA Version: 10.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  GeForce GTX 108...  Off  | 00000000:02:00.0 Off |                  N/A |
|  0%   50C    P8    15W / 280W |    284MiB / 11178MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0     18694      C   ...s/miniconda3/envs/cil2_demos/bin/python   137MiB |
|    0     19798      C   ...s/miniconda3/envs/cil2_demos/bin/python   137MiB |
+-----------------------------------------------------------------------------+
```

paskino:
The solution is achieved by following these steps:


  - https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#why-do-i-get-the-error-unknown-runtime-specified-nvidia
  - https://github.com/nvidia/nvidia-container-runtime#installation
  - https://nvidia.github.io/nvidia-container-runtime/

KrisThielemans:
@paskino found that the docker image hasn't been built with GPU support. Confirmed at
https://travis-ci.com/github/SyneRBI/SIRF-SuperBuild/jobs/506993066#L4458

Possibly a travis or CMake issue as this seems fine
[...]

PR #551, opened 2021-06-03
[docker] add BASE_IMAGE to GPU image

KrisThielemans:
addresses #547

### DIFF
--- a/docker/docker-compose.srv-gpu.yml
+++ b/docker/docker-compose.srv-gpu.yml
@@ -6,6 +6,7 @@ services:
   build:
    target: service
    args:
+    BASE_IMAGE: nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04
     EXTRA_BUILD_FLAGS: "-DBUILD_CIL=ON"
    cache_from:
     - synerbi/sirf:core-gpu


KrisThielemans:
We use
https://github.com/SyneRBI/SIRF-SuperBuild/blob/c4394bca9884f61ac4b9640c4d0db978b033f0b9/docker/Dockerfile#L1-L2
`ARG` variables can be overriden at build-time, see https://docs.docker.com/engine/reference/builder/#arg.

My current understanding is that we build GPU always via `docker-compose.srv-gpu.yml`
https://github.com/SyneRBI/SIRF-SuperBuild/blob/c4394bca9884f61ac4b9640c4d0db978b033f0b9/.travis.yml#L122
This file defines 3 "services": `sirf`, `core` and `gadgetron`. These are all independent, although `sirf` happens to cache from `sirf/core-gpu`.

`core` (which will create an image called `core-gpu`) specifies the `BASE_IMAGE`
https://github.com/SyneRBI/SIRF-SuperBuild/blob/c4394bca9884f61ac4b9640c4d0db978b033f0b9/docker/docker-compose.srv-gpu.yml#L17
but `sirf` did not. It is not sufficient for an image to `cache-from` to inherit its `BASE` (indeed, `cache-from` is designed such that you can have different bases, but still benefit from the caching of a few layers).

Let's see what happens when I merge it I guess...

PR #558, opened 2021-06-06
docker: no longer chown /opt

### LINKED ISSUE #555: docker: slow startup of container the very first time
If you get a new image and start it up, the very first time it can take a long time. Our doc says "a few minutes" (used to say "a few seconds"). For me, it easily takes 15 minutes, but this is docker running inside the VM (which isn't really recommended).

I did the following to check what's going on.
```
./sirf-compose-server up -d sirf
docker exec -w / ps aux
```
repeating the latter a few times.

It turns out it is busy in https://github.com/SyneRBI/SIRF-SuperBuild/blob/32b0126e00dd69b15b3223fb1b8334e2d5804bcf/docker/entrypoint.sh#L34-L39
The longest part seems to be when updating `/opt/pyvenv` (which is where we install miniconda (see [here](https://github.com/SyneRBI/SIRF-SuperBuild/blob/6d0db321393e13aa6428dea808e46a30a74f3545/docker/user_python-ubuntu.sh#L40) and [here](https://github.com/SyneRBI/SIRF-SuperBuild/blob/6d0db321393e13aa6428dea808e46a30a74f3545/docker/user_python-ubuntu.sh#L12)), although it takes it time with `/opt/SIRF-Superbuild` (which is the build directory) as well of course.

Any ideas to speed this up?

(It seems non-trivial to avoid this, as docker file permissions need to know who the "host" user is, see #464)

KrisThielemans:
At least on some systems, changing ownership of /opt/* took a long time when the container was created. Instead, we just give everyone write permissions, but do this when creating the images.

Fixes #555

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,7 +147,7 @@ jobs:
         $DCC build sirf
         # extract updated ccache
         # sudo rm -rf devel/.ccache/*
-        $DCC run --rm sirf /bin/bash -c 'sudo rm -rf /devel/.ccache/* && sudo cp -a /opt/ccache/* /devel/.ccache/'
+        $DCC run --rm sirf /bin/bash -c 'rm -rf /devel/.ccache/* && cp -a /opt/ccache/* /devel/.ccache/'
         # replace travis' ccache with the built images's
         mv devel/.ccache ~
       ); fi
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,11 @@
 
 ## v3.x.x
 - disable built of NiftyPET by default as requires Python2 for which we dropped support
-- docker: add dependencies that are available only from conda-forge (for CIL)
+- docker:
+  - major change w.r.t. users and permissions. We know build as user jovyan (by default)
+    and still switch to sirfuser for running the container. This avoids having to
+    reset permissions of many files, and therefore speeds-up container start-up.
+  - add dependencies that are available only from conda-forge (for CIL)
 - updated ISMRMRD to 1.4.2.1
 
 ## v3.0.0
--- a/docker/DocForDevelopers.md
+++ b/docker/DocForDevelopers.md
@@ -26,6 +26,11 @@ We use the [`ENTRYPOINT` mechanism](https://docs.docker.com/engine/reference/bui
 
 The very first time the container is run, `entrypoint.sh` creates the `sirfuser` user, copies files in correct places, and takes care of file permissions (see below).
 
+### Why the jovyan user?
+The `NB_USER` (by default called `jovyan`) is a convention used by JupyterHub. It launches jupyter
+notebooks under this user. Due to permission problems, it should not have `root` access.
+We build all files as this user.
+
 ### File permissions
 
 Quoting from https://blog.gougousis.net/file-permissions-the-painful-side-of-docker/.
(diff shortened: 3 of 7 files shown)


KrisThielemans:
container start-up time is now indeed very fast, while creation time is very long...

We should try to be more selective probably, and definitely not do the whole of `/opt` again at the end of the `service`  image creation (where it should only have to do whatever was installed as extra  in `/opt/pyenv`)

KrisThielemans:
Ah well. I thought I had it fixed with my new strategy: use `umask 000` when calling `user_*sh`. Indeed, this fixes permissions of `/opt/pyenv` and `/opt/SIRF-SuperBuild` as I expected to be read/writable for everyone:
```
$ ls -l /opt/
total 24
drwxr-xr-x 2 root root 4096 Jun  5 17:20 bin
drwxrwxrwx 1 root root 4096 Jun  7 21:16 ccache
lrwxrwxrwx 1 root root   25 Jun  5 17:20 cmake -> cmake-3.13.4-Linux-x86_64
drwxr-xr-x 6 root root 4096 Jun  5 17:20 cmake-3.13.4-Linux-x86_64
drwxrwxrwx 1 root root 4096 Jun  7 22:02 pyvenv
drwxrwxrwx 1 root root 4096 Jun  7 22:02 SIRF-SuperBuild
```
Result: image creation time same as on `master`, but container creation time 1 second! Yay!

Except... it doesn't work. Files in /opt/SIRF-SuperBuild/sources/SIRF` and `.../builds/SIRF/build` (and all others) are not writable by `sirfuser`, and a `conda install coverage` fails with a permission problem.

It seems that setting `umask 000` gets ignored by same programs. Oh well.

KrisThielemans:
So, the question is then if we can live with `sirfuser` not being able to write to most of `/opt`. Consequences
- `conda install `fails
- `sudo conda install` fails as it's not in the path (fixable)
- `sudo /opt/pyvenv/bin/conda install` seems to work
- `pip install` works
- updating the `/opt/SIRF-SuperBuild` would need `sudo`

Presumably it would be safest to `chmod .../sources/SIRF/data`

Of course, this would have to be documented.

It's not ideal, but shaving 10 mins of start-time isn't small...

Or is there another solution?

paskino:
Is it not possible to use `sirfuser` to run some stuff, like building SuperBuild, install conda etc?

There's no reason why conda should be installed by the superuser.

KrisThielemans:
> Is it not possible to use `sirfuser` to run some stuff, like building SuperBuild, install conda etc?

The point is that at image build time (i.e. whatever `Dockerfile` does), you don't know yet what the UID:GID has to be. You could do it with 1000:1000 and hope to be lucky, but we'd still have to cope with other values.

(Note that the username/groupname don't mean anything, it's the numbers that matter).

KrisThielemans:
Haroon Chughtai suggests
> another option could be to create and grant permissions to the needed folders to a group during the build. Then when the image is pulled and run, the created user just needs to be added to that group to have the right access? 

I guess this might work as:
- Dockerfile
    - create group (with crazy gid)
    - (I think) create user (with crazy uid)
    - run the `user_sh` as this user
 - `entrypoint.sh`
    - add `sirfuser` to the above group
 
First check in a VM (or running docker) that the above strategy actually allows `sirfuser` to write to the relevant directories.

KrisThielemans:
@paskino I've taken some of your lines in #569 and incorporated them in the "regular" stuff. I've kept @casperdcl's overall structure, but do make another user (part of `users` group), and make `sirfuser` also part of `users`, as suggested by Haroon. This way, we don't need to `chown` most files anymore. Image build time is a bit longer (as it still has to do some `chown`), but container build time is then very fast.

We have therefore 2 users: `jovyan` (builds) and `sirfuser` (runs and mounts `/devel`).

However, `conda` is run by `jovyan`, while `bash` (in `sirf:latest`) or `jupyter` (in `service`) are run by `sirfuser` , which could/will lead to confusion.

Note that with this set-up, it should be easy (?) to copy some of your lines at the end of the current `Dockerfile` for a new `jupyterhub` service.

I hope this was clear enough, but probably not!

KrisThielemans:
@paskino I tried this together with #574 It seems to work quite well. Reasonably fast build time (some `chown` stuff still), and very fast container start-up. I see no down-sides (aside from the extra complication...).

It should also be almost ready for the jupyterhub changes.

I believe we should merge these PRs. (There's a conflict, which I'll sort out).

KrisThielemans:
@paskino I've cleaned up the history and merged it. Let's see what the Travis build says for the generated dockers...

PR #550, opened 2021-06-03
disable niftypet

paskino:
closes #549

PR #561, opened 2021-06-07
disable NiftyPET by default as it requires python2

paskino:
PR #550 disabled NiftyPET only for docker builds, however, since `NiftyPET` requires Python 2 and we dropped support for Python2 NiftyPET should just be disabled throughout and enabled by a daring user.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 # ChangeLog
 
 ## v3.x.x
-- no longer build NiftyPET on docker until we have transitioned it to Python3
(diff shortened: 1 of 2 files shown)


ashgillman:
Latest NiftyPET is Py3, but perhaps best to leave disable while there are no active user of NiftyPET via SIRF?

https://github.com/NiftyPET/NiftyPET/pull/5

KrisThielemans:
yeah. we need to check if STIR works with NiftyPET3 and adapt STIR. Feel free to do that... but not now!

PR #560, opened 2021-06-07
correct .dockerignore file

paskino:
closes #534

PR #564, opened 2021-06-08
562 conda forge deps

### LINKED ISSUE #562: docker: when to install CIL prerequisites
Part of CIL prerequisites are listed in https://github.com/SyneRBI/SIRF-SuperBuild/blob/0e4a5eb575bdf80ff4c3fdf8643777f2f941e11d/docker/requirements.txt#L4-L11 however, 2 are missing: `PyWavelets` and `dxchange` which are not available on pip but are on `conda-forge`.

Given that the docker image is using conda python I suggest to change
https://github.com/SyneRBI/SIRF-SuperBuild/blob/0e4a5eb575bdf80ff4c3fdf8643777f2f941e11d/docker/user_python-ubuntu.sh#L35-L43

as
```bash 


# Python (runtime)

if [ "$PYTHON" = "miniconda" ]; then
    conda install --yes --file requirements.txt  
  fi
  conda update -c conda-forge -y --all
  conda clean -y --all
else
    pip install -U -r requirements.txt
  fi
fi

 

```

paskino:
closes #562

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## v3.x.x
 - disable built of NiftyPET by default as requires Python2 for which we dropped support
+- docker: add dependencies that are available only from conda-forge (for CIL)
 
 ## v3.0.0
 - travis to use BUILD_CIL=ON for all Docker builds
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -1,3 +1,3 @@
 devel/**
-!devel/.ccache/**
+!devel/.ccache
 devel/.ccache/.gitignore
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -69,9 +69,10 @@ RUN chmod 644 .bashrc .profile
 
 # Python (virtualenv)
 COPY requirements.txt .
+COPY requirements_conda_forge.txt .
 COPY user_python-ubuntu.sh .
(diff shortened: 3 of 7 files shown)


paskino:
from https://github.com/SyneRBI/SIRF-SuperBuild/pull/560#issuecomment-856531673 

@KrisThielemans says:
> the requirements_conda_forge.txt seems incomplete. You would need to list all other packages as well as far as I can see.

All packages can be installed via conda except https://github.com/SyneRBI/SIRF-SuperBuild/blob/3f4cc7294aa37d49c151d958e53474a9d91fae7f/docker/requirements.txt#L3 which `conda-build` doesn't know how to treat.

So I chose to separate the install requirements into the original one that can be installed by `pip` and the one where the package does not exist in `pip` but only exists in `conda-forge` (like CIL).

https://github.com/paskino/SIRF-SuperBuild/blob/fa6c53c3199f679663f54f5858f31336549d29b5/docker/user_python-ubuntu.sh#L36-L46

KrisThielemans (approved):

KrisThielemans:
ok. thanks for the clarification. I didn't check that's what you're doing, but I believe you!

paskino:
Fantastic! GitHub Actions are having issues https://www.githubstatus.com/

KrisThielemans:
merge anyway.

PR #574, opened 2021-06-14
Docker small fixes

KrisThielemans:
see commit messages if what this does.

@paskino, could you have a quick look?

### DIFF
--- a/docker/build_system-ubuntu.sh
+++ b/docker/build_system-ubuntu.sh
@@ -3,7 +3,11 @@ set -ev
 # SIRF external dependencies
 apt-get update -qq
 apt-get install -yq --no-install-recommends \
-  libboost-all-dev     \
+  libboost-dev libboost-chrono-dev \
+  libboost-filesystem-dev libboost-thread-dev \
+  libboost-date-time-dev libboost-regex-dev \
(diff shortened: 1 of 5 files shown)


docker/build_system-ubuntu.sh
```diff
@@ -3,7 +3,11 @@ set -ev
 # SIRF external dependencies
 apt-get update -qq
 apt-get install -yq --no-install-recommends \
-  libboost-all-dev     \
+  libboost-dev libboost-chrono-dev \
+  libboost-filesystem-dev libboost-thread-dev \
+  libboost-date-time-dev libboost-regex-dev \
+  libboost-program-options-dev libboost-atomic-dev \
+  libboost-test-dev libboost-timer-dev \
```
paskino, line 10: didn't we decide to remove these packages?

paskino (approved):
this should be fine

docker/build_system-ubuntu.sh
KrisThielemans: I copied the list from the VM. Not 100% sure if we still need all of them, but definitely most of them.

The previous version installed all the boost libraries, which is considerably more

KrisThielemans:
thanks. feel free to merge (*not* squash-merge), or tell me to do it

PR #575, opened 2021-06-14
use setup.py in installation

paskino:


KrisThielemans:
ready to merge?

KrisThielemans:
I see one job was cancelled but the others were fine. could merge `master` again if you feel it is necessary

PR #581, opened 2021-06-16
pass to SIRF if Gadgetron used CUDA

paskino:
If Gadgetron is built with CUDA, SIRF needs to pass the include to CUDA during compilation of RPE encoding code.


KrisThielemans:
You will need to update the `SIRF_TAG` presumably, once https://github.com/SyneRBI/SIRF/pull/951 is merged

paskino:
Absolutely

AnderBiguri (approved):
Works well on my side (compiles correctly).

PR #585, opened 2021-06-16
bump SIRF version

paskino:


KrisThielemans:
can you commit this with [travis skip] please? It's going to fail but first waste 3 hours of computation

PR #586, opened 2021-06-16
Update .bashrc

### LINKED ISSUE #573: docker (and others): Gadgetron or ISRMRD can link with hdf5 but not find it at run-time
In our docker, we have 2 hdf5 libraries: one installed with `apt` and the other one with `conda`. They don't have the same version of course.

During build time, it uses the one from `conda` (possibly because it is more recent). However, this is not in our `LD_LIBRARY_PATH`.
Therefore gadgetron fails. See https://github.com/SyneRBI/SIRF-SuperBuild/pull/566#issuecomment-859718511.

I would have thought that a `conda activate base` would add `/opt/pyvenv/lib`, but it doesn't. Maybe I do that wrong?

In any case, it seems to imply that we need to add the location of the `hdf5` library to `LD_LIBRARY_PATH` somehow.

paskino:
[travis skip]
closes #573

### DIFF
--- a/docker/.bashrc
+++ b/docker/.bashrc
@@ -26,7 +26,7 @@ export PS1='sirf$ '
 
 # Need to add this as we have built using some of these shared libraries
 # See https://github.com/SyneRBI/SIRF-SuperBuild/issues/573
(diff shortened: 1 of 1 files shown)


KrisThielemans:
Turns out this isn't necessary and the original code works fine

PR #593, opened 2021-06-17
Docker improvements

### LINKED ISSUE #572: docker: access to SIRF-Contrib from jupyter notebook
as it starts in `/devel` , the user isn't able to (easily) see the source of the contribs, while they are used in the exercises.

Note that currently the SIRF-SuperBuild installs them in `/opt/SIRF-SuperBuild/INSTALL/python/sirf/contribs` on the docker image.

### LINKED ISSUE #589: docker: need to freeze SIRF-SuperBuild tag
currently we're building `master`. Needs to be fixed for a particular version possibly? I think it will be alright for the `release` versions on DockerHub, but if people build it themselves, they will currently get the latest, which would be surprising.

KrisThielemans:
See commit messages

Fixes #589 
Fixes #572 

but also some introduces a `NUM_PARALLEL_BUILDS` variable

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,8 @@
   - major change w.r.t. users and permissions. We know build as user jovyan (by default)
     and still switch to sirfuser for running the container. This avoids having to
     reset permissions of many files, and therefore speeds-up container start-up.
+  - introduced `SIRF_SB_URL`, `SIRF_SB_TAG` and `NUM_PARALLEL_BUILDS`. They default to the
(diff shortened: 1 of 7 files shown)


ashgillman:
A shame about the COPY - but yes as I mentioned I recalled having some issues there. Perhaps we can still have a long-term goal issue for it?

Otherwise, looks good to me

KrisThielemans:
> Perhaps we can still have a long-term goal issue for it?

yes, please create it

PR #598, opened 2021-06-22
bump CIL tag

paskino:


KrisThielemans (approved):
sure. However, why don't we bump the SIRF_TAG as well to get us a bit closer. Can do it in one PR as far as I'm concerned (less Travis time)

paskino:
Sure, let's wait until the 3.1.0 tag is done.

PR #607, opened 2021-06-28
docker: create user with gid

### LINKED ISSUE #606: docker: entrypoint fails if group exists
Found by @mastergari

`./sirf-compose` tries to find the group of the current user, and create a group in the docker image accordingly. However, when that group already exists, subsequent `adduser` fails:
```
“Creating sirfuser:501:20
addgroup: The GID `20' is already in use.
adduser: The group `sirfuser' does not exist.
addgroup: The user `sirfuser' does not exist.
Updating file ownership for /home/sirfuser
chown: invalid user: ‘sirfuser:sirfuser’
Switching to sirfuser and executing /usr/local/bin/service.sh
error: failed switching to "sirfuser": unable to find user sirfuser: no matching entries in passwd file”
```

KrisThielemans:
The entrypoint.sh failed if `GROUP_ID` refers to an existing group, as the `adduser` command referred to it via the wrong name.

Fixes #606

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -31,7 +31,7 @@ addgroup --quiet --system --gid "$GROUP_ID" "$mainUser"
 #   -p $(echo somepassword | openssl passwd -1 -stdin)
 adduser --quiet --system --shell /bin/bash \
   --no-create-home --home /home/"$mainUser" \
-  --ingroup "$mainUser" --uid "$USER_ID" "$mainUser"
+  --gid "$GROUP_ID" --uid "$USER_ID" "$mainUser"
 addgroup "$mainUser" users
 
 echo "$mainUser ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers.d/"$mainUser"
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,9 @@
 # ChangeLog
 ## v3.x.x
 
-- added external project astra-python-wrapper to allow updates of ASTRA without conflicts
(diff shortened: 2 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -31,7 +31,7 @@ addgroup --quiet --system --gid "$GROUP_ID" "$mainUser"
 #   -p $(echo somepassword | openssl passwd -1 -stdin)
 adduser --quiet --system --shell /bin/bash \
   --no-create-home --home /home/"$mainUser" \
-  --ingroup "$mainUser" --uid "$USER_ID" "$mainUser"
+  --gid "$GROUP_ID" --uid "$USER_ID" "$mainUser"
```
paskino, line 34: `man adduser` reports 

> Users' primary groups can also be overridden from the command line with the `--gid` or  `--ingroup` options  to  set  the group by id or name, respectively.

Given that the group is created with the `${GROUP_ID}` with name `${mainUser}`, i.e. `sirfuser`, what would the update you made really change?
https://github.com/SyneRBI/SIRF-SuperBuild/blob/ff8dc6544320ca82a58e275f0b7490808d647410/docker/entrypoint.sh#L28

paskino (approved):
I built it locally and it works for me (however that's on linux)

docker/entrypoint.sh
KrisThielemans: it shouldn't change anything! *except* in the case where the group already exists. In that case, `--gid` will be fine, but `--ingroup` won't, as it won't be called `$mainUser`

PR #614, opened 2021-08-10
declare behaviour with policy 0077

paskino:
declare what behaviour we expect


KrisThielemans:
We currently do not depend on the choice of the policy (i.e. we never create a local variable that is then used in `option`), but this way we are sure about the behaviour of CMake.

PR #628, opened 2021-11-18
fix corner cases for Boost

### LINKED ISSUE #627: system boost found even when building our own and other boost problems
# USE_SYSTEM_BOOST=OFF

On Ubuntu 20.04, if boost is installed via `apt`, but we set `USE_SYSTEM_BOOST=OFF`, CMake (in dependent projects such as ISMRMD) finds the system boost anyway. This is despite us passing `BOOST_ROOT` etc https://github.com/SyneRBI/SIRF-SuperBuild/blob/448356272bf150846115daeafe55a2e600c1a1cd/SuperBuild/External_Boost.cmake#L80-L84

I think what's going on is that Ubuntu 20.04 uses Boost 1.71.0 which comes with `BoostConfig.cmake`. (We currently build Boost 1.65.0 which doesn't install `BoostConfig.cmake`). Presumably [`FindBoost.cmake`](https://cmake.org/cmake/help/latest/module/FindBoost.html) tries to find a "config" version first before even looking at `BOOST_ROOT`.

The result is that some compilations could use our installed `boost` include files, while the linking will happen against the system version. This leads to linking problems like this
```
/usr/bin/ld: CMakeFiles/test_ismrmrd.dir/test_acquisitions.cpp.o: in function `__static_initialization_and_destruction_0(int, int) [clone .constprop.141]':
test_acquisitions.cpp:(.text.startup+0xa1): undefined reference to `boost::unit_test::ut_detail::auto_test_unit_registrar::auto_test_unit_registrar(boost::unit_test::basic_cstring<char const>, boost::unit_test::basic_cstring<char const>, unsigned long, boost::unit_test::decorator::collector&)'
```
or run-time crashes.

This was first observed by @NicoleJurjew.

# USE_SYSTEM_BOOST=ON
In this case we want to pass the location on, and do
https://github.com/SyneRBI/SIRF-SuperBuild/blob/448356272bf150846115daeafe55a2e600c1a1cd/SuperBuild/External_Boost.cmake#L94-L96

however, CMake allows another variable `BOOSTROOT` for the same purpose, as well as `Boost_DIR` to find a "config" version. We should therefore set those as well.

KrisThielemans:
- when USE_SYSTEM_BOOST=OFF, set Boost_NO_SYSTEM_PATHS=ON to increase our chances of finding our own version
- when USE_SYSTEM_BOOST=OFF, pass BOOSTROOT and Boost_DIR to projects (when set)

Fixes #627

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 
 - moved the VM repository to the `VirtualBox` subdirectory
 - fix usage of `proj_EXTRA_CMAKE_ARGS` facility (it was broken for all projects except ITK) [#616](https://github.com/SyneRBI/SIRF-SuperBuild/issues/616)
+- Boost: fix cases where the wrong version of boost could be found [#627](https://github.com/SyneRBI/SIRF-SuperBuild/issues/627)
 
 ## v3.1.1
 
--- a/SuperBuild/External_Boost.cmake
+++ b/SuperBuild/External_Boost.cmake
@@ -1,7 +1,7 @@
 #========================================================================
 # Author: Benjamin A Thomas
 # Author: Kris Thielemans
-# Copyright 2017, 2020 University College London
+# Copyright 2017, 2020, 2021 University College London
 #
 # This file is part of the CCP SyneRBI (formerly PETMR) Synergistic Image Reconstruction Framework (SIRF) SuperBuild.
 #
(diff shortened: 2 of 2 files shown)


KrisThielemans:
One job fails due to unrelated https://github.com/TomographicImaging/CIL/issues/1083

KrisThielemans:
@paskino, any objections for merging?

paskino:
If I understand correctly the failure is not there anymore?

KrisThielemans:
CIL failures? Sure, GHA seems happy

paskino:
Yes, CIL failures. So I will close https://github.com/TomographicImaging/CIL/issues/1083

PR #630, opened 2021-11-29
Add the CERN ROOT library (used by STIR)

### LINKED ISSUE #21: dis/enable ROOT support in STIR by default
provide option to enable for advanced users

KrisThielemans:
Introduce `USE_ROOT` (defaulting to `OFF`) variable to build/use the CERN ROOT library, and pass it to STIR.

Fixes #21 (by making it obsolete)


KrisThielemans:
@paskino can you help with the following?

We currently build ROOT without support for X11. However, this means that the `root` executable isn't built, see https://sft.its.cern.ch/jira/browse/ROOT-10948

We therefore need a post-ROOT-install step that says: update `ROOT_INSTALL_DIR/bin/root` from `ROOT_INSTALL_DIR/bin/root.exe` (i.e. copy `root.exe` if `root` doesn't exist or if it is older).

paskino:
Looking at ROOT's code it seems that `x11` is anyhow required to build.

https://github.com/root-project/root/blob/4398ac49ab78db96354450f5f5a78ffb2a58fc3d/cmake/modules/SearchInstalledSoftware.cmake#L335-L375

KrisThielemans:
well, they might say that in that code, but note that it is within `if (x11)`. You can build the whole thing with setting that to `OFF`. That will disable some stuff and speed-up compilation. (Of course, X11 isn't used on Windows).

KrisThielemans:
Maybe you're saying not to bother and just build with x11 on. Sound simpler so fine for me

KrisThielemans:
ROOT-related failure which should be fixed now that when https://github.com/UCL/STIR/pull/971 has been merged. Re-running jobs therefore

paskino:
> Maybe you're saying not to bother and just build with x11 on. Sound simpler so fine for me

I was trying to gather information as I didn't understand the proposed solution. To me it looked that on Linux it would build only if X11 was found. Why would we have `root.exe` then is a mystery to me.

KrisThielemans:
Slightly less enthusiastic now about building with `x11=ON`, as we then [need](https://root.cern/install/dependencies/#required-packages-1) libx11-dev, libxpm-dev, libxft-dev and libxext-dev.

I'd hate to install these all the time, but at some point, we might need it, e.g. on the VM for people who like GATE, or of course for testing on GHA.

What about making a `docker/build-ROOT.sh` containing
```
# APT dependencies for building ROOT
# Note: X libraries would not be required when building ROOT with `X11=OFF`. However,
# this does not build the `root` executable (even though `root.exe` can be used instead).
sudo apt install libssl-dev libx11-dev libxpm-dev libxft-dev libxext-dev
```
and modifying the README to say that if you want to build ROOT, that file can be used to install dependencies.

paskino:
Sure. Can you explain to me if and why would we have `root.exe`?

KrisThielemans:
good point. We currently only use it directly in the testing. https://github.com/UCL/STIR/blob/master/recon_test_pack/run_root_GATE.sh

I guess I also could put in a change there and see if `root` doesn't exist, then try `root.exe`. That's maybe the simplest fix.

KrisThielemans:
Using `root.exe` now in the STIR tests: https://github.com/UCL/STIR/commit/4c0b7637cb33ea7d11699c084b160f433ee96f45, so this should be good to go

KrisThielemans:
hmmm. no GHA on this merge? No idea why. I guess I'll forge ahead...

Issue #444, opened 2020-11-17
need to update ITK version


KrisThielemans:
we're currently on ITK 4.13, but that doesn't support gcc-9 (see #443). Let's attempt to upgrade to v5.1.1

KrisThielemans:
we're currently at 4.13.1. 4.13.3 is available, but it'd be better to get to ITK 5. A few things that stop us:
- [x] STIR had a problem with ITK 5.1, but fixed with https://github.com/UCL/STIR/pull/791
- [ ] ITK 5.0 doesn't build on Windows, 5.1.2 doesn't build with an external HDF5. (Fixed in https://github.com/InsightSoftwareConsortium/ITK/pull/2247, now merged to ITK master)
- [ ] PETPVC doesn't support ITK 5 yet, https://github.com/UCL/PETPVC/issues/55. We're currently not building PETPVC yet, but it'd be nice to have.

KrisThielemans:
@paskino please try with ITK 5.0.1, but keep 4.13.3 on Windows for a while longer.

KrisThielemans:
time to try ITK 5.2.1...

PR #632, opened 2022-01-05
ITK 5.2.1

KrisThielemans:
Upgrade ITK. I had to disable the MINC part of the IO library as it failed to compile with Ubuntu's HDF5 library:
```
The selected version of HDF5 library does not support setting backwards compatibility at run-time.\
  Please use a different version of HDF5
```
generated at https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/MINC/src/libminc/libsrc2/volume.c#L39-L40

An alternative would be to set `ITK_USE_SYSTEM_HDF5=OFF`, at the expense of larger ITK build. I have introduced an option to be able to do this.

@ashgillman any opinions on this?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 - moved the VM repository to the `VirtualBox` subdirectory
 - fix usage of `proj_EXTRA_CMAKE_ARGS` facility (it was broken for all projects except ITK) [#616](https://github.com/SyneRBI/SIRF-SuperBuild/issues/616)
 - Boost: fix cases where the wrong version of boost could be found [#627](https://github.com/SyneRBI/SIRF-SuperBuild/issues/627)
+- updated versions:
+  - ITK: 5.2.1
 
 ## v3.1.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@
 - fix usage of `proj_EXTRA_CMAKE_ARGS` facility (it was broken for all projects except ITK) [#616](https://github.com/SyneRBI/SIRF-SuperBuild/issues/616)
 - Boost: fix cases where the wrong version of boost could be found [#627](https://github.com/SyneRBI/SIRF-SuperBuild/issues/627)
 - updated versions:
-  - ITK: 5.2.1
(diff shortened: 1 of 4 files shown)


KrisThielemans:
fixes #444, see there for extra info

KrisThielemans:
Failure as STIR [cannot find `itkOrientImageFilter`](https://github.com/SyneRBI/SIRF-SuperBuild/runs/4714035924?check_suite_focus=true#step:6:8162). This apparently [sits in ITKImageGrid](https://github.com/InsightSoftwareConsortium/ITK/blob/57ef2ac9f5fe97799aefb69192bb265f4d0ff391/Modules/Filtering/ImageGrid/include/itkOrientImageFilter.h#L138)

ashgillman:
> An alternative would be to set ITK_USE_SYSTEM_HDF5=OFF, at the expense of larger ITK build. I have introduced an option to be able to do this.

Would it perhaps be better to use pass HDF5_DIR to ITK "if not USE_SYSTEM_HDF5" and use that if there are issues rather than deferring to ITK to install an additional, different HDF5?

ashgillman:
I'm not sure why its still failing w/ ITKImageGrid...

KrisThielemans:
> Would it perhaps be better to use pass HDF5_DIR to ITK "if not USE_SYSTEM_HDF5" and use that if there are issues rather than deferring to ITK to install an additional, different HDF5?

This is the default situation with `ITK_USE_SYSTEM_HDF5=ON`. However, the problem is that to remain compatible with MATLAB/Python, we sometimes need to build/use an HDF5 version that is incompatible with ITK...  As far as I understand, if we set `ITK_USE_SYSTEM_HDF5=OFF`, ITK will indeed build its own (subset of) HDF5 , but avoid conflicts by renaming all functions/class with an itk prefix, see some info here https://github.com/InsightSoftwareConsortium/ITK/blob/9ff84420e55d2f86984f40f14861bddf0812b82b/Modules/ThirdParty/HDF5/src/itkhdf5/src/itk_hdf5_mangle.h. Those functions will only be used by ITK, not by SIRF/STIR/ISMRMRD/Gadgetron.

KrisThielemans:
All done. I'm fine to merge...

PR #634, opened 2022-01-05
ISMRMRD 1.5.0

### LINKED ISSUE #633: ISMRMRD failures on Windows
we've switched to ISMRMRD 1.4.2.1. This gives 2 problems on Windows:
- boost linking errors in https://github.com/ismrmrd/ismrmrd/pull/170
- a linking error fixed in https://github.com/ismrmrd/ismrmrd/pull/165

On Windows we have to use 1.5.0 therefore. It compiles fine for me, but test currently fail probably because of my local set-up (and attempts to use https://github.com/SyneRBI/SIRF/pull/936)

@evgueni-ovtchinnikov any idea if this would work ?

KrisThielemans:
fixes #633

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@
 - moved the VM repository to the `VirtualBox` subdirectory
 - fix usage of `proj_EXTRA_CMAKE_ARGS` facility (it was broken for all projects except ITK) [#616](https://github.com/SyneRBI/SIRF-SuperBuild/issues/616)
 - Boost: fix cases where the wrong version of boost could be found [#627](https://github.com/SyneRBI/SIRF-SuperBuild/issues/627)
+- updated versions:
+  - ISMRMRD: 1.5.0
 
 ## v3.1.1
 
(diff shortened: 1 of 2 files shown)


KrisThielemans:
@evgueni-ovtchinnikov do you want to check this before merging?

evgueni-ovtchinnikov:
building ISMRMRD with SuperBuild gives me this puzzling error:
```
pathspec 'v1.5.0' did not match any file(s) known to git
```

KrisThielemans:
This could be due to `ISMRMRD_URL` still pointing to our fork, or because your CMake is too old (they had a bug related to fetching tags which is fixed now).

In case of the latter, you could just delete `builds/ISMRMRD/build` and `sources/ISMRMRD` and try again.

evgueni-ovtchinnikov:
actually, the error was elsewhere - I did not notice that URL was wrong (`https://github.com/SyneRBI/ismrmrd` instead of `https://github.com/ismrmrd/ismrmrd`)

all builds ok now and all Gadgetron tests passed

PR #641, opened 2022-01-21
[GHA] rearrange matrix and add gcc9 and USE_SYSTEM*=OFF tests

KrisThielemans:


KrisThielemans:
various problems are coming up when building dependencies (of course): #644, #643,#642. I'll switch those off again

Issue #569, opened 2021-06-12
docker image for kubernetes jupyterhub


KrisThielemans:
We're trying to get an image ready to launch on jupyterhub with kubernetes. Current attempt is on https://github.com/paskino/SIRF-SuperBuild/tree/add_jovyan_user/docker

This does not use `docker-compose`, so our `.yml` files are not used.

Difficulties
- [x] `sirf_env.sh` is not sourced
- [x] the jupyter notebook started in `/home/jovyan`
- [ ] `gadgetron` is not started (but we can tell the user to do that, as they need to do it on the VM anyway)
- [x] user permissions for `/opt` (read-only) - see related discussion in #464, #555 and #558 although they concentrate on using `docker-dompose` -  causing problems (solvable by doing `sudo chown -R jovyan:users /opt/SIRF-Exercises/`)
   - would not be able to modify notebooks
   - cannot download data
      ```
      $ /opt/SIRF-Exercises/scripts/download_data.sh -p ~/data ~/data_download
      ....
      creating data_path.py in /opt/SIRF-Exercises/lib/sirf_exercises/data_path.py
      ./download_data.sh: line 204: /opt/SIRF-Exercises/lib/sirf_exercises/data_path.py: Permission denied
      ```
 - [x] <strike>currently starting JupyterLab, where'd prefer to use the classic interface (can be done by adding `tree` at the URL)</strike> (This is a jupyterlab (system-wide!) setting, not under our control
 - [x] <strike>cannot plot anything</strike> (This disappears when launching classic jupyter)
  ```
  plt.figure()
  plt.plot(1,2)
  Javascript Error: IPython is not defined
  [<matplotlib.lines.Line2D at 0x7f46d0e47580>]
  ```

paskino:
I built the image with docker-compose. The problem is how the image is run, which we don't know. I believe that with a lengthy docker terminal command one can achieve the same result as with `docker-compose`. 

I can make such that the directory where we clone the notebooks are `rw` for the user `jovyan`.

KrisThielemans:
some information on the set-up.

- it uses [KubeSpawner](https://jupyterhub-kubespawner.readthedocs.io/en/latest/)
- Information on configuration of JupyterHub:
https://zero-to-jupyterhub.readthedocs.io/en/latest/jupyterhub/customizing/user-environment.html

KrisThielemans:
[...]

PR #650, opened 2022-02-02
Update the Jupyterhub docker build

paskino:
This will close https://github.com/SyneRBI/SIRF-SuperBuild/issues/569 using a different tactic, explained at https://github.com/paskino/SIRF-SuperBuild/blob/jupyterhub_docker_build/jupyterhub/README.md

The strategy is basically to:
  1. modify the `datascience-notebook` to have the `nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04` base image, `paskino/jupyter:datascience-notebook-cuda10-ubuntu18.04`
  1. build the `synerbi/sirf:sirf-core` image with the `nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04` base image
  1. build the jupyterhub image from the image at point 1, copy the SIRF `INSTALL` directory from the `synerbi/sirf:sirf-core` (previous step), set the appropriate environmental variable and install CIL via conda

I suggest that we separate the `requirements.txt` in the `docker` directory for the ones that are SIRF and CIL specific. In this way I could use the `requirements.txt` instead of hard coding the requirements in the [Dockerfile](https://github.com/paskino/SIRF-SuperBuild/blob/52b968e09acd2e16e67da10b521964ccaba6947a/jupyterhub/Dockerfile#L53-L54)

I have run a few notebooks in the STFC cloud, both SIRF and CIL and they work. Ah, no, connection to Gadgetron doesn't work...

```
02-02 16:53:23.324 DEBUG [gadgetron_home.cpp:98] Gadgetron home: "/opt/SIRF-SuperBuild/INSTALL"

02-02 16:53:23.324 INFO [GadgetStreamController.cpp:48] Connection from 127.0.0.1:41296
02-02 16:53:23.324 INFO [GadgetStreamController.cpp:226] Found 1 readers
02-02 16:53:23.324 INFO [GadgetStreamController.cpp:227] Found 1 writers
02-02 16:53:23.324 INFO [GadgetStreamController.cpp:228] Found 4 gadgets
02-02 16:53:23.325 INFO [GadgetStreamController.cpp:244] --Found reader declaration
02-02 16:53:23.325 INFO [GadgetStreamController.cpp:245]   Reader dll: gadgetron_mricore
02-02 16:53:23.325 INFO [GadgetStreamController.cpp:246]   Reader class: GadgetIsmrmrdAcquisitionMessageReader
02-02 16:53:23.325 INFO [GadgetStreamController.cpp:247]   Reader slot: 1008
02-02 16:53:23.327 ERROR [GadgetStreamInterface.h:57] Failed to load DLL, Possible reasons: 
02-02 16:53:23.327 ERROR [GadgetStreamInterface.h:58]    * Name of DLL is wrong in XML file 
02-02 16:53:23.327 ERROR [GadgetStreamInterface.h:59]    * Path of DLL is not in your DLL search path (LD_LIBRARY_PATH on Unix)
02-02 16:53:23.327 ERROR [GadgetStreamInterface.h:60]    * Path of other DLLs that this DLL depends on is not in the search path
02-02 16:53:23.327 ERROR [GadgetStreamController.cpp:254] Failed to load GadgetMessageReader from DLL
02-02 16:53:23.327 ERROR [GadgetStreamController.cpp:124] GadgetStream configuration failed
02-02 16:53:23.327 DEBUG [GadgetStreamController.cpp:159] handle_close called
02-02 16:53:23.327 INFO [GadgetStreamController.cpp:164] Shutting down stream and closing up shop...
02-02 16:53:23.328 DEBUG [EndGadget.cpp:12] Close called in EndGadget with flags 1
02-02 16:53:23.328 DEBUG [EndGadget.cpp:23] Calling close in base class  with flags 1
02-02 16:53:23.328 DEBUG [Gadget.h:174] Gadget (EndGadget) Close Called with flags = 1
02-02 16:53:23.328 DEBUG [Gadget.h:184] Gadget (EndGadget) waiting for thread to finish
02-02 16:53:23.328 DEBUG [EndGadget.cpp:12] Close called in EndGadget with flags 0
02-02 16:53:23.328 DEBUG [EndGadget.cpp:23] Calling close in base class  with flags 0
02-02 16:53:23.328 DEBUG [Gadget.h:174] Gadget (EndGadget) Close Called with flags = 0
02-02 16:53:23.328 DEBUG [Gadget.h:186] Gadget (EndGadget) thread finished
02-02 16:53:23.328 DEBUG [Gadget.h:123] Shutting down Gadget (EndGadget)
02-02 16:53:23.328 INFO [GadgetStreamController.cpp:190] Stream is closed
^C02-02 16:53:32.924 DEBUG [GadgetServerAcceptor.cpp:82] GadgetServerAcceptor::handle_close
02-02 16:53:32.924 DEBUG [GadgetServerAcceptor.cpp:83] Close Data Acceptor
jovyan@jupyter-sirf-2dedoardo:~$ ldd gadgetron
ldd: ./gadgetron: No such file or directory
jovyan@jupyter-sirf-2dedoardo:~$ ldd $(which gadgetron)
        linux-vdso.so.1 (0x00007ffeeffdd000)
        libgadgetron_gadgetbase.so.3.17 => /opt/SIRF-SuperBuild/INSTALL/lib/libgadgetron_gadgetbase.so.3.17 (0x00007f2a8131b000)
        libgadgetron_toolbox_rest.so.3.17 => /opt/SIRF-SuperBuild/INSTALL/lib/libgadgetron_toolbox_rest.so.3.17 (0x00007f2a810c2000)
        libgadgetron_toolbox_cloudbus.so.3.17 => /opt/SIRF-SuperBuild/INSTALL/lib/libgadgetron_toolbox_cloudbus.so.3.17 (0x00007f2a80e9f000)
        libACE-6.4.5.so => /usr/lib/libACE-6.4.5.so (0x00007f2a80b0d000)
        libgadgetron_toolbox_log.so.3.17 => /opt/SIRF-SuperBuild/INSTALL/lib/libgadgetron_toolbox_log.so.3.17 (0x00007f2a80907000)
        libboost_filesystem.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.65.1 (0x00007f2a806ed000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f2a80364000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2a8014c000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2a7ff2d000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2a7fb3c000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2a7f79e000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f2a8179e000)
        libgadgetron_toolbox_node_discovery.so.3.17 => /opt/SIRF-SuperBuild/INSTALL/lib/libgadgetron_toolbox_node_discovery.so.3.17 (0x00007f2a7f579000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2a7f375000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f2a7f16d000)
        libboost_system.so.1.65.1 => /usr/lib/x86_64-linux-gnu/libboost_system.so.1.65.1 (0x00007f2a7ef68000)

```


KrisThielemans:
I'm a bit confused by the many references to stuff on `paskino` branches, also in comments, but maybe this is pending a clean-up, or maybe I need to spend a bit more time on it 😄

paskino:
Yes, the plan would be to add a GHA to create the jupyterhub image from this and host the jupyter datascience-notebook image in synerbi (?).

paskino:
The problem was due to a different location of conda from our docker image (`/opt/pyvenv`) and the jupyter docker stack (`/opt/conda`). 

To control where we install conda, I added a build argument `PYTHON_INSTALL_DIR` 
https://github.com/paskino/SIRF-SuperBuild/blob/68b94f8e2dd5495c2322f3fac57492124603c374/docker/Dockerfile#L75 which is then used during the install of conda (or of a virtual environment).

https://github.com/paskino/SIRF-SuperBuild/blob/68b94f8e2dd5495c2322f3fac57492124603c374/docker/user_python-ubuntu.sh#L4

The last bits to fix:
 - [ ] the `.bashrc` where the location of conda/virtual environmet is hard coded (bug).
 - [ ] the docker compose files which need to specify the location of conda

Although we seem to support both a virtual environment and conda, in reality we only use the latter. Therefore, I suggest to remove the additional complication of supporting virtual environment, and install conda in `/opt/conda`, i.e. remove [these](https://github.com/paskino/SIRF-SuperBuild/blob/68b94f8e2dd5495c2322f3fac57492124603c374/docker/user_python-ubuntu.sh#L20-L28) lines. 

Besides, this will ensure that we can use the latest `core` image to install SIRF on the jupyterhub image, without the need of using non-default location for conda: currently `/opt/pyvenv`. 

@KrisThielemans any suggestions?

KrisThielemans:
Hi

I certainly got confused by the `conda` vs `pip` options in the Docker stuff. But then I am confused by `conda` vs `pip` in general. I would have thought that a venv would work with a conda Python, but maybe it's not optimal.  think `conda`'s version checking is a bit more stringent than `pip`, as conda recommends creating a conda environment specifying all packages, but certainly not sure). Here's the related tip from [their doc](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-with-commands)
> Install all the programs that you want in this environment at the same time. Installing 1 program at a time can lead to dependency conflicts.

I am a little bit reluctant to do different things in our dockerfile than what we recommend to users, but as long as our GHA tests the "pure" `pip` install route, that should be fine.

After this ramble, I think that a simplification of the Docker set-up is good.

KrisThielemans:
By the way, why are the 1804 jobs failing? Possibly unrelated to this PR

paskino:
I haven't understood why the 18.04 builds fail but it is related to the conda vs pip issue as it happens when running `user_python-ubuntu.sh`. The (not) funny thing is that, on my machine the 20.04 fails while the 18.04 works!

I'm going to remove the pip stuff and default to conda in the docker stuff!

paskino:
> I am a little bit reluctant to do different things in our dockerfile than what we recommend to users, but as long as our GHA tests the "pure" `pip` install route, that should be fine.

Our GHA does not test the "pure" `pip` option. Actually there's no way it can as the install of conda/virtualenv is done with [default parameters](https://github.com/SyneRBI/SIRF-SuperBuild/blob/3c8679bda27af6d8ab1495029e97a5bb47300757/docker/Dockerfile#L102) (i.e. install `conda`) and no mechanism to override them is in place.

KrisThielemans:
> Our GHA does not test the "pure" pip option

not with Docker, but the "normal" one. https://github.com/SyneRBI/SIRF-SuperBuild/blob/3c8679bda27af6d8ab1495029e97a5bb47300757/.github/workflows/c-cpp.yml#L78

KrisThielemans:
so cannot remove `pip`. I really don't want to force everyone to use `conda`. (I find it a bit of a mess, certainly on Windows)

paskino:
Thanks for spotting that!

paskino:
> > Our GHA does not test the "pure" pip option
> 
> not with Docker, but the "normal" one.
> 
> https://github.com/SyneRBI/SIRF-SuperBuild/blob/3c8679bda27af6d8ab1495029e97a5bb47300757/.github/workflows/c-cpp.yml#L78

Ah! that's probably the reason why the 18.04 builds fail! They use the stock python that is installed in Ubuntu 18.04 (py36) which is too old. Although the solution the error message mentions, is actually the line that it is running.

BTW, I never said it explicitly, but the new jupyterhub image works on the STFC cloud, with SIRF 3.1.0 and CIL 21.3.0!!!

docker/user_python-ubuntu.sh
```diff
@@ -19,7 +19,9 @@ miniconda)
   ;;
 *python*)
   # virtualenv
-  curl https://bootstrap.pypa.io/get-pip.py > get-pip.py
+  
+  # curl https://bootstrap.pypa.io/get-pip.py > get-pip.py
+  curl https://bootstrap.pypa.io/pip/3.6/get-pip.py > get-pip.py
```
paskino, line 24: temporary fix for GHA failing on Ubuntu 18.04. 

How do I go about checking the python version and download the right file without hard coding it?

KrisThielemans:
> They use the stock python that is installed in Ubuntu 18.04 (py36) which is too old. 

why too old?

KrisThielemans:
oh, sorry. for the `get-pip` stuff? You'd have hoped they check this themselves. oh well. I'll have to leave you to it...

paskino:
> > They use the stock python that is installed in Ubuntu 18.04 (py36) which is too old.
> 
> why too old?

The error line states
`ERROR: This script does not work on Python 3.6 The minimum supported Python version is 3.7. Please use https://bootstrap.pypa.io/pip/3.6/get-pip.py instead.`

docker/user_python-ubuntu.sh
paskino: I added script to do this

paskino:
> > fix .bashrc file which has hard coded the location of conda/virtualenv
> 
> hardcoding is fine in a docker image?
> 

In practice we now give the user the possibility to choose where to put their conda/virtualenv, with the build argument `PYTHON_INSTALL_DIR` but we hardcode the location in the `.bashrc`. Notice that **only** GHA do not use the standard choice (conda in /opt/conda), as you pointed out
https://github.com/SyneRBI/SIRF-SuperBuild/blob/3c8679bda27af6d8ab1495029e97a5bb47300757/.github/workflows/c-cpp.yml#L78

In my opinion the `.bashrc` should also be generated with the right directory. I'm unsure how to do this. 

> > decide the default python on the docker image we push to dockerhub (currently conda)
> 
> I don't see why we wouldn't use `conda` Python. Other we cannot use `conda` at all (loads of conflicts)
> 
Perfect

> > decide whether not to build CIL by default in the docker image and install it via conda (adding an additional step probably).
> 
> maybe you meant "or"? I'm happy with a `conda install` of CIL

Fine for me. It's going to be easier for us. Then, only GHA are going to continue building CIL as part of the SuperBuild.

paskino:
I'm wondering how to install CIL via conda. The thing is:

1. the SuperBuild allows build of CIL + (some) dependencies
2. Some dependencies of CIL are installed via [conda](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker/requirements_conda_forge.txt). Such dependencies are not related to the CIL functionality that is used in SIRF
2. GHA do not install conda, nor does it seem to rely on it
3. Docker installs conda by default

In docker, we have some level of control over this with the build arguments, but I think it's insufficient:
* `PYTHON_INSTALL_DIR` controls where the conda/virtualenv is installed
* `EXTRA_BUILD_FLAGS` with which we can set whether to build CIL as part of the SuperBuild

I suppose there should be some additional build argument.
One that allows to modify the `PYTHON` variable in
 https://github.com/SyneRBI/SIRF-SuperBuild/blob/3c8679bda27af6d8ab1495029e97a5bb47300757/docker/user_python-ubuntu.sh#L5 

Another one that allows the user to specify whether to install CIL via conda or SuperBuild, like `CIL_INSTALL_METHOD`. 

Of course, in the case of conda install one should also make sure that `-DBUILD_CIL=OFF`  is set. However, this is very complicated to achieve in a Dockerfile, so I'd suggest to rely on **documentation**.

KrisThielemans:
I'm really quite confused by all this. In your docker set-up, you decide how to build CIL/SIRF. you hardwire `.bashrc` etc accordingly. What kind of "user" do you have in mind that will want to change `Dockerfile` properties somehow?

It's maybe different if we want to let people use the `user_python-ubuntu.sh` etc on their own Linux system (or our VM). Then more configuration options are necessary. But that doesn't have anything to do with a docker build?

I'm really quite confused by all this. 😜

paskino:
GHA do not behave as a normal user. They install virtualenv not conda, so there is no possibility to install CIL with conda there. 

On GHA we have a configuration which builds CIL. So now, the same Dockerfile needs to support 2 ways of installing CIL. Or am I mistaken?

Next, GHA does install virtualenv in a non default directory, which is not the one hard coded in `.bashrc`.

KrisThielemans:
GHA runs whatever is on the workflow, which doesn't use `Dockerfile` (but it does happen to use some files in `docker/`). It doesn't use `.bashrc` (only a docker instance does).

`Dockerfile` doesn't care about the GHA workflow. It uses some files in `docker`.

Both can therefore decide to install CIL in entirely different ways. No?

paskino:
Oooo you are right!

CHANGES.md
```diff
@@ -1,6 +1,8 @@
 # ChangeLog
 ## vx.x.x
 
+- Docker images install CIL via conda, so setting `BUILD_CIL=OFF` in SuperBuild
```
KrisThielemans, line 4:
```suggestion
- Docker images install CIL via conda, so setting `BUILD_CIL=OFF` in the docker build
```

version_config.cmake
```diff
@@ -151,9 +151,9 @@ set(DEFAULT_JSON_TAG v3.9.1)
 
 # CCPi CIL
 set(DEFAULT_CIL_URL https://github.com/TomographicImaging/CIL.git)
-set(DEFAULT_CIL_TAG "v21.2.0")
+set(DEFAULT_CIL_TAG "v21.3.1")
 set(DEFAULT_CIL-ASTRA_URL https://github.com/TomographicImaging/CIL-ASTRA.git)
-set(DEFAULT_CIL-ASTRA_TAG "v21.2.0")
+set(DEFAULT_CIL-ASTRA_TAG "v21.3.0")
```
KrisThielemans, line 156: are you sure about the space here?

jupyterhub/README.md
```diff
@@ -0,0 +1,121 @@
+## Build the Jupyterhub image 
+
+To create the image with SIRF and all the other stuff required by jupyterhub we start from the `datascience-notebook` from https://github.com/jupyter/docker-stacks.
+
+However, we require GPU access (for [CIL](https://github.com/TomographicImaging/CIL.git)).
+
+The strategy is:
+  1. to modify the `datascience-notebook` to have the `nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04` base image, `paskino/jupyter:datascience-notebook-cuda10-ubuntu18.04`
```
KrisThielemans, line 8: add a link to your github fork/branch here

KrisThielemans (changes requested):
great stuff. minor things only

jupyterhub/README.md
paskino: done. added sentence 
> A few mods for use with Ubuntu 18.04 are in the fork https://github.com/paskino/docker-stacks/tree/base_image_ubuntu18.04 which is used as base for creating the new `datascience-notebook`.

KrisThielemans:
don't wait for me to merge this. do it when you feel it's ready

PR #692, opened 2022-04-19
Fix python dest bashrc

### LINKED ISSUE #691: Allow use of parameters of user_python-ubuntu.sh
There is a conflict between the requirements for jupyterhub and docker and the files in the docker directory used by GHA:

1. by default docker installs miniconda in[ `/opt/pyvenv` ](https://github.com/SyneRBI/SIRF-SuperBuild/blob/fcea2e436551802326716a0420052ee171f22c2c/docker/user_python-ubuntu.sh#L4-L5) as [ `Dockerfile` ](https://github.com/SyneRBI/SIRF-SuperBuild/blob/fcea2e436551802326716a0420052ee171f22c2c/docker/Dockerfile#L103)does not set the parameters 
2. `.bashrc` hardcodes `/opt/conda` as location for conda python, which would be OK for jupyterhub but not for standard docker. But in the end is good for neither  https://github.com/SyneRBI/SIRF-SuperBuild/blob/fcea2e436551802326716a0420052ee171f22c2c/docker/.bashrc#L21-L29
3. GHA use user_python-ubuntu.sh from command line and use system `python3` and `~/virtualenv` directory for the install of SIRF
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fcea2e436551802326716a0420052ee171f22c2c/.github/workflows/c-cpp.yml#L78 

The proposed solution is to call `user_python-ubuntu.sh` with parameters in `Dockerfile`, and to remove the hard coded path in `.bashrc`.

paskino:
closes #691

### DIFF
--- a/docker/user_python-ubuntu.sh
+++ b/docker/user_python-ubuntu.sh
@@ -1,8 +1,27 @@
 #!/usr/bin/env bash
+
+# User can pass 2 parameters to this script: 
+# 1. PYTHON_EXECUTABLE which can be either the 'miniconda' string or the location of the Python executable
+# 2. PYTHON_INSTALL_DIR which points to the location where the user has their virtual environment installed
+#    The virtual environment can be either virtualenv or conda
+#
+# To pass these parameters the user needs to set (at least) temporary environment variables:
+# PYTHON_EXECUTABLE=python3 PYTHON_INSTALL_DIR="~/virtualenv" bash user_python-ubuntu.sh
+# Fx`xor details see https://github.com/SyneRBI/SIRF-SuperBuild/pull/692#issuecomment-1102704682 
 [ -f .bashrc ] && . .bashrc
 set -exv
-INSTALL_DIR="${1:-/opt/conda}"
-PYTHON="${2:-miniconda}"
+
+if [ -n "${PYTHON_EXECUTABLE}" ]; then 
+  PYTHON=$PYTHON_EXECUTABLE
+else
+  PYTHON='miniconda'
+fi
+
+if [[ -n "${PYTHON_INSTALL_DIR}" ]]; then
+  INSTALL_DIR=${PYTHON_INSTALL_DIR}
+else
+  INSTALL_DIR="/opt/conda"
+fi
 
 # Python
 case "$PYTHON" in
--- a/docker/user_python-ubuntu.sh
+++ b/docker/user_python-ubuntu.sh
@@ -7,7 +7,7 @@
 #
 # To pass these parameters the user needs to set (at least) temporary environment variables:
 # PYTHON_EXECUTABLE=python3 PYTHON_INSTALL_DIR="~/virtualenv" bash user_python-ubuntu.sh
-# Fx`xor details see https://github.com/SyneRBI/SIRF-SuperBuild/pull/692#issuecomment-1102704682 
+# For details see https://github.com/SyneRBI/SIRF-SuperBuild/pull/692#issuecomment-1102704682 
 [ -f .bashrc ] && . .bashrc
 set -exv
 
--- a/docker/user_python-ubuntu.sh
+++ b/docker/user_python-ubuntu.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 [ -f .bashrc ] && . .bashrc
 set -exv
-INSTALL_DIR="${1:-/opt/pyvenv}"
+INSTALL_DIR="${1:-/opt/conda}"
 PYTHON="${2:-miniconda}"
 
 # Python
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -75,7 +75,7 @@ jobs:
           sudo bash build_python-ubuntu.sh;
           sudo bash build_gadgetron-ubuntu.sh;
           sudo bash build_system-ubuntu.sh;
-          bash user_python-ubuntu.sh ~/virtualenv python3 ;
+          PYTHON_EXECUTABLE=python3 PYTHON_INSTALL_DIR=~/virtualenv bash user_python-ubuntu.sh;
     - name: ccache
       uses: hendrikmuhs/ccache-action@v1
       with:
--- a/docker/.bashrc
+++ b/docker/.bashrc
@@ -18,7 +18,7 @@
 # play safe with an innocent prompt that also works in the Bourne shell
 export PS1='sirf$ '
 # Python (virtualenv)
-[ -f /opt/conda/bin/activate ] && . /opt/conda/bin/activate
+[ -f PYTHON_INSTALL_DIR/bin/activate ] && . PYTHON_INSTALL_DIR/bin/activate
 
 # SIRF env
 [ -f /opt/SIRF-SuperBuild/INSTALL/bin/env_sirf.sh ] && \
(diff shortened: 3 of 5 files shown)


paskino:
Additional problem: the call

https://github.com/SyneRBI/SIRF-SuperBuild/blob/fcea2e436551802326716a0420052ee171f22c2c/docker/user_python-ubuntu.sh#L14


calls  the following code:

```bash
#!/bin/sh
_CONDA_ROOT="/opt/pyvenv"
# Copyright (C) 2012 Anaconda, Inc
# SPDX-License-Identifier: BSD-3-Clause
\. "$_CONDA_ROOT/etc/profile.d/conda.sh" || return $?
conda activate "$@"
```
In this case `$@` will expand to the parameters passed to `user_python-ubuntu.sh`. Unfortunately this is not correct as activate will only work if passed a name of an environment and certainly not if passed 2 parameters, which return 

```
conda activate "$@"
++ conda activate /opt/conda miniconda
++ local cmd=activate
++ case "$cmd" in
++ __conda_activate activate /opt/conda miniconda
++ '[' -n '' ']'
++ local ask_conda
+++ PS1='sirf$ '
+++ __conda_exe shell.posix activate /opt/conda miniconda
+++ __add_sys_prefix_to_path
+++ '[' -n '' ']'
++++ dirname /opt/conda/bin/conda
+++ SYSP=/opt/conda/bin
++++ dirname /opt/conda/bin
+++ SYSP=/opt/conda
+++ '[' -n '' ']'
+++ PATH=/opt/conda/bin:/opt/conda/condabin:/usr/lib/ccache:/opt/cmake/bin:/usr/                                                         local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/u                                                         sr/bin:/sbin:/bin:/home/jovyan/.local/bin
+++ export PATH
+++ /opt/conda/bin/conda shell.posix activate /opt/conda miniconda
activate does not accept more than one argument:
['/opt/conda', 'miniconda']
```

paskino:
Any idea on how to achieve this @KrisThielemans @ashgillman ?

Additionally, I was thinking to pass these info as temporary environment variables:

`PYTHON_EXECUTABLE='sa' PYTHON_INSTALL_DIR="/opt/pyvenv" bash test_my_vars.sh`

with `test_my_vars.sh` as 

```bash

if [ -n PYTHON_EXECUTABLE ]; then 
  PYTHON='miniconda'
else
  PYTHON=$PYTHON_EXECUTABLE
fi

if [ -n PYTHON_INSTALL_DIR ]; then
  INSTALL_DIR="/opt/conda"
else
  INSTALL_DIR=${PYTHON_INSTALL_DIR}
fi

echo "PYTHON_EXECUTABLE ${PYTHON}"
echo "PYTHON_INSTALL_DIR ${INSTALL_DIR}"
```

but my output is 
```
PYTHON_EXECUTABLE miniconda
PYTHON_INSTALL_DIR /opt/conda
```
while I expected

```
PYTHON_EXECUTABLE sa
PYTHON_INSTALL_DIR /opt/pyvenv
```

KrisThielemans:
The `activate` problem.
==============
Not sure why they have `$@` in a script to be sourced. Seems to make no sense to me. What's in `"$_CONDA_ROOT/etc/profile.d/conda.sh" `?
I suppose you could cheat and just replace the `source .../activate` line with the 2 lines that it actually contains (but removing the `$@`).

The environment problem
=================
I believe you need (probably with the quotes, not 100% sure)
```sh
if [ -n "$PYTHON_EXECUTABLE" ]; then
...
```

paskino:
Thanks, the environment problem is due to my poor knowledge of bash! I missed a `$` and inverted the test case! It should be

```bash 

if [ -n "${PYTHON_EXECUTABLE}" ]; then 
  PYTHON=$PYTHON_EXECUTABLE
else
  PYTHON='miniconda'
fi

if [[ -n "${PYTHON_INSTALL_DIR}" ]]; then
  INSTALL_DIR=${PYTHON_INSTALL_DIR}
else
  INSTALL_DIR="/opt/conda"
fi

echo "PYTHON_EXECUTABLE ${PYTHON}"
echo "PYTHON_INSTALL_DIR ${INSTALL_DIR}"
```

Double quotes are not necessary.

KrisThielemans:
> Double quotes are not necessary.

are you sure? suppose the path contains a space?

paskino:
You are right, they were not necessary to fix my problem, but they are needed to prevent the problem of a path with a space!

paskino:
Some checks fail on [building CIL](https://github.com/SyneRBI/SIRF-SuperBuild/runs/6095484623?check_suite_focus=true#step:7:2055), however other configurations build CIL, both in devel and no devel config.

The error unveils something weird: `cmake` is being passed/passing the wrong path. Apparently it thinks that the `CMAKE_CURRENT_BUILD_DIRECTORY` is `/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/sources/CIL/Wrappers/Python/` which doesn't make sense.

```
CMake Error at Wrappers/Python/cmake_install.cmake:54 (file):
  file INSTALL cannot find
  "/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/sources/CIL/Wrappers/Python/build/lib/cil":
  No such file or directory.
  cmake_install.cmake:48 (include)
```

docker/user_python-ubuntu.sh
```diff
@@ -1,8 +1,27 @@
 #!/usr/bin/env bash
+
+# User can pass 2 parameters to this script: 
+# 1. PYTHON_EXECUTABLE which can be either the 'miniconda' string or the location of the Python executable
+# 2. PYTHON_INSTALL_DIR which points to the location where the user has their virtual environment installed
+#    The virtual environment can be either virtualenv or conda
+#
+# To pass these parameters the user needs to set (at least) temporary environment variables:
+# PYTHON_EXECUTABLE=python3 PYTHON_INSTALL_DIR="~/virtualenv" bash user_python-ubuntu.sh
+# Fx`xor details see https://github.com/SyneRBI/SIRF-SuperBuild/pull/692#issuecomment-1102704682
```
KrisThielemans, line 10:
```suggestion
# For details see https://github.com/SyneRBI/SIRF-SuperBuild/pull/692#issuecomment-1102704682 
```

KrisThielemans (changes requested):
typo

KrisThielemans:
> The error unveils something weird: `cmake` is being passed/passing the wrong path. Apparently it thinks that the `CMAKE_CURRENT_BUILD_DIRECTORY` is `/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/sources/CIL/Wrappers/Python/` which doesn't make sense.

can you point to the CIL CMake file? The SB one has the line https://github.com/SyneRBI/SIRF-SuperBuild/blob/fcea2e436551802326716a0420052ee171f22c2c/SuperBuild/External_CIL.cmake#L63
I cannot see anything wrong with that (nor related to the error)

paskino:
Sure, this is for the non devel build. https://github.com/TomographicImaging/CIL/blob/f5c06566215564ad78ba0f1738d0836f3ad37a60/Wrappers/Python/CMakeLists.txt#L99-L100

KrisThielemans:
I note https://github.com/TomographicImaging/CIL/blob/f5c06566215564ad78ba0f1738d0836f3ad37a60/Wrappers/Python/CMakeLists.txt#L81-L82 but this is for `WIN32`, and is therefore presumably irrelevant. (Can't say I understand why you'd want to do that for WIN32, but it's up to you).

In any case, this seems unrelated to this PR, so ok to merge from my side

paskino:
Oh yes, that's a remnant of some previous code. In fact the install command is exactly the same for `WIN32` or not.

paskino:
ah no! that seems to be a bug https://github.com/TomographicImaging/CIL/issues/1226

PR #693, opened 2022-04-25
Ismrmrd min required

### LINKED ISSUE #635: Set minimum ISMRMRD required version to 1.4.2
Once https://github.com/SyneRBI/SIRF/pull/936 is merged, during the build of SIRF we will require ISMRMRD version 1.4.2.

paskino:
closes #635 

sets required versions for STIR and NIFTYREG


SuperBuild.cmake
```diff
@@ -198,6 +198,17 @@ option(USE_SYSTEM_NIFTYREG "Build using an external version of NIFTYREG" OFF)
 option(USE_SYSTEM_GTest "Build using an external version of GTest" OFF)
 option(USE_SYSTEM_ACE "Build using an external version of ACE" ON)
 
+if (USE_SYSTEM_STIR)
+  set(STIR_REQUIRED_VERSION "4.1.0" CACHE STRING ON)
+endif()
```
KrisThielemans, line 203:
```suggestion
set(STIR_REQUIRED_VERSION "4.1.0" CACHE STRING ON)
```
nice and simple. However, these lines should go into `version_config.cmake`

SuperBuild.cmake
```diff
@@ -198,6 +198,17 @@ option(USE_SYSTEM_NIFTYREG "Build using an external version of NIFTYREG" OFF)
...
+
+
+if (USE_SYSTEM_siemens_to_ismrmrd)
```
KrisThielemans, line 206: probably needs to be
```cmake
if (BUILD_siemens_to_ismrmrd)
```
also should be in `version_config.cmake`

SuperBuild.cmake
```diff
@@ -277,6 +288,7 @@ endif()
 
 # parallelproj
 set(USE_parallelproj ON CACHE BOOL "Build STIR with parallelproj's projectors") # FORCE)
+set(parallelproj_REQUIRED_VERSION "0.8" CACHE STRING ON)
```
KrisThielemans, line 291: also should be in `version_config.cmake`

KrisThielemans (changes requested):

SuperBuild.cmake
paskino: I believe `BUILD_siemens_to_ismrmrd = NOT USE_SYSTEM_siemens_to_ismrmrd`.

SuperBuild.cmake
paskino: Of course we can always add the `${proj}_REQUIRED_VERSION` which will be used only in the case we use the system `${proj}`

SuperBuild.cmake
KrisThielemans: no, not as far as I can see. In fact, `USE_SYSTEM_siemens_to_ismrmrd` is nowhere used. Even stronger, it should not exist as a `USE_SYSTEM*` variable only makes sense for "library" type of projects. I would remove it.

KrisThielemans (commented):
Should the `*REQUIRED_VERSION` be cached variables? These are not "user-settable", except in somewhat contrived cases (example: someone sets e.g .`SIRF_TAG` to a lower version, *and* is using a system ISMRMRD version that is older than what would be required).

We could keep it cached, but then I would certainly make them `ADVANCED`. But I'd prefer not to have them cached: otherwise we'd have one more set of variables to delete when upgrading .

version_config.cmake
```diff
@@ -179,10 +179,16 @@ if (DEVEL_BUILD)
...
+  set(STIR_REQUIRED_VERSION "4.1.0")
 
   ## siemens_to_ismrmrd
   set(DEFAULT_siemens_to_ismrmrd_URL https://github.com/ismrmrd/siemens_to_ismrmrd )
   set(DEFAULT_siemens_to_ismrmrd_TAG b87759e49e53dab4939147eb52b7a0e6465f3d04)
+  if (USE_SYSTEM_siemens_to_ismrmrd)
+    set(ISMRMRD_REQUIRED_VERSION "1.4.2.1")
+  else()
+    set(ISMRMRD_REQUIRED_VERSION "1.7")
+  endif()
```
KrisThielemans, line 191:
```suggestion
  if (BUILD_siemens_to_ismrmrd)
    set(ISMRMRD_REQUIRED_VERSION "1.7")
  else()
    set(ISMRMRD_REQUIRED_VERSION "1.4.2.1")
  endif()
```

KrisThielemans:
Build failures is probably due to https://github.com/TomographicImaging/CIL/issues/1226

KrisThielemans:
If you're happy with my suggestion, please squash-merge.

SuperBuild.cmake
paskino: Ah, all right

paskino:
I had not noticed the existence of `BUILD_siemens_to_ismrmrd` but did notice `USE_SYSTEM_siemens_to_ismrmrd`. 

CIL bugfix is [going to be merged](https://github.com/TomographicImaging/CIL/pull/1227) and I will bump the version/commit as soon as it is merged.

PR #694, opened 2022-04-25
Added CITATION.cff file

DANAJK:


paskino:
I'd merge this.

KrisThielemans:
See comments in https://github.com/SyneRBI/SIRF/pull/1091

KrisThielemans:
Also, possibly confusingly, the authorship of the SB as currently recorded in this `zenodo.json` is different from the one of `SIRF`. I think that is how it should be. This implies adapting this `CITATION.cff`

KrisThielemans:
Looks like we do need https://github.com/citation-file-format/cffconvert-github-action. Possibly it can be combined with some smart rules for when to run it (i.e. only when `CITATION.cff` is updated) as well as not run other actions when only `CITATION.cff` is updated, but we could leave that for later...

DANAJK:
GitHub has checked the CITATION.cff file and found a mistake in my recent manual edit. I have corrected it in the most recent commit. Github will also give you a bibtex file so I think we are all done, except for doing a pull request. (On GitHub, switch to the Zenodo-citation branch and you will see on the right a link to "Cite this repository"). The Zenodo community might need manually updating, but that seems easier than manually syncing a zenodo.json and CITATION.cff file.

PR #703, opened 2022-05-09
[ci skip] updates the path in the update_VM.sh script

paskino:
The script `update_VM.sh` refers to the archived repo https://github.com/SyneRBI/SyneRBI_VM 
https://github.com/SyneRBI/SIRF-SuperBuild/blob/775d4a18bccc0bc67b194622eb976acd61620a8f/VirtualBox/scripts/update_VM.sh#L4-L6

I updated the location of the `UPDATE.sh` script.

### DIFF
--- a/VirtualBox/scripts/update_VM.sh
+++ b/VirtualBox/scripts/update_VM.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 echo "Updating..."
-cd ~/devel/SyneRBI_VM
+cd ~/devel/SIRF-SuperBuild/VirtualBox
 git pull
 cd scripts
 


evgueni-ovtchinnikov:
so far as I remember, the idea was that `update_VM.sh` could be run from any location, so it must be in `~/devel/install/bin`, or else `~/devel/SIRF-SuperBuild/VirtualBox/scripts` must be added to `$PATH`, is it not?

evgueni-ovtchinnikov:
@KrisThielemans you seem to be right, as Edoardo and I are having problems installing CIL on VM 3.2.0, which prevent me from running some Jupyter notebooks.

paskino:
The problem of installing CIL on the VM is very recent and due to a change in the install script of CIL which we have just addressed https://github.com/TomographicImaging/CIL/pull/1247 and it has nothing to do with the problem with the `update_VM.sh`

As far as I can see the lines pointed out by Kris are never used in the new VM as they try to handle the rename of the repo from  CCPPETMR_VM to SyneRBI_VM.

@evgueni-ovtchinnikov you need to get CIL master to use that fix and `update_VM.sh` will never give it to you.

evgueni-ovtchinnikov:
@paskino but `UPDATE.sh` is run by `update_VM.sh` ;)

> @evgueni-ovtchinnikov you need to get CIL master

You mean I need to set CIL_TAG to `origin/master`? Same for CIL-ASTRA_TAG?

KrisThielemans:
I guess `UPDATE.sh` will only get CIL master once `version_config.cmake` would ask for it, *and* we tag the SIRF-Superbuild again. Before doing the latter, you should be able to run `update_VM.sh -t origin/master` or something like that (once the version is updated)

evgueni-ovtchinnikov:
KT> Maybe that's what @evgueni-ovtchinnikov is trying to do.

Exactly. `Introductory/acquisition_model_mr_pet_ct.ipynb` failed because it needed `cil.plugins.astra`, which my VM 3.2.0 did not have, so I decided to update it, but VM had old `update_VM.sh` in `~/devel/install/bin`.

KrisThielemans:
ok, so our `update_VM.sh` forwards to `UPDATE.sh` trick fails in this case. sigh.

Feels to me that we have 3 options;
- tell them they cannot update their VM (I guess I veto)
- tell them they need to `git pull SIRF-SuperBuild; cp .../update_VM.sh ~/devel/install/bin` (not ideal)
- unfreeze SynerRBI_VM repo, change its `UPDATE.sh` to do the above, exit, and tell them to re-run
-

paskino:
Can't we do 
> tell them they need to `git pull SIRF-SuperBuild; cp .../update_VM.sh ~/devel/install/bin` (not ideal)

in `UPDATE.sh`?

KrisThielemans:
> Can't we do
> 
> > tell them they need to `git pull SIRF-SuperBuild; cp .../update_VM.sh ~/devel/install/bin` (not ideal)
> 
> in `UPDATE.sh`?

well, it has to be in `SyneRBI_VM/scripts/UPDATE.sh` as that's the one that will be executed. Can then just as well do it there (and cut out all the rest). no?

paskino:
Yes, yes, I was actually writing

Ah, no, because they'd have an old `update_VM.sh` script. So if you veto solution 1 and find solution 2 not ideal, the only option is 3.

Of course I'd be for 1. :P

KrisThielemans:
I think option 3 might be easy. But if not, we can do 2 (or the half-way solution in between and just write a message)

KrisThielemans:
@paskino we should merge, and create for the update to old VM repo (or just do it 😄 )

paskino:
Yes, I'll do it ASAP

PR #704, opened 2022-05-10
Pass location of ipp library to CIL CMake

paskino:
Build of CIL would not pick the installed version of IPP.


KrisThielemans:
a bit ugly to remove a test that fails 😄  but I agree with doing that for now such that we can move on. I didn't check what the other failure is. As far as I'm concerned, ok to merge.

KrisThielemans (approved):
merge when done

Issue #709, opened 2022-05-16
Docker: setting LD_LIBRARY_PATH breaks library resolution


paskino:
We are struggling to resolve the appropriate libraries at runtime, see #573 or https://github.com/SyneRBI/SIRF/issues/1094 and all devel builds are failing.

The current solution for docker is to add to `LD_LIBRARY_PATH` in `.bashrc`
https://github.com/SyneRBI/SIRF-SuperBuild/blob/1024422a85c54d480d1a33a441c13fd9f862c2c1/docker/.bashrc#L27-L29

However, during a docker build I stumbled upon a huge problem: having set `LD_LIBRARY_PATH` makes so that some libraries that are in the conda install are picked before the system ones, so that basic tools, like `mv`, don't work and the build fails. BTW @KrisThielemans had already [suspected](https://github.com/SyneRBI/SIRF-SuperBuild/issues/573#issuecomment-860175624) this!

I was thinking to use `RPATH` for Gadgetron. It seems they already use the [`INSTALL_RPATH`](https://github.com/gadgetron/gadgetron/blob/9e733b193c60df78575577a237d1c237f0eb3826/CMakeLists.txt#L109)

```cmake
set(CMAKE_INSTALL_RPATH ".:${CMAKE_INSTALL_PREFIX}/lib")
set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
```

But they don't use the [`BUILD_RPATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_RPATH.html).

paskino:
To me it looks like that this is a problem only if `miniconda` is selected at 
https://github.com/SyneRBI/SIRF-SuperBuild/blob/1024422a85c54d480d1a33a441c13fd9f862c2c1/docker/Dockerfile#L75

because also might install several other _system_ libraries.

KrisThielemans:
I believe this has been resolved. Maybe some stuff to do for jupyterhub. If so, @paskino reassign to 3.4

KrisThielemans:
@paskino I think this needs to be closed.

PR #711, opened 2022-05-16
Fix gadgetron link

paskino:
As described in #709 there are problem resolving libraries at runtime and at build time. This PR tries to address this.


paskino:
I'm not sure about the `.bashrc`. I think that's a problem but maybe it has nothing to do with this PR.

docker/user_sirf-ubuntu.sh
```diff
@@ -18,6 +18,4 @@ echo $COMPILER_FLAGS
 echo $BUILD_FLAGS $EXTRA_BUILD_FLAGS
 cmake $BUILD_FLAGS $EXTRA_BUILD_FLAGS $COMPILER_FLAGS .
 
-cmake --build . -j ${NUM_PARALLEL_BUILDS} --target Gadgetron
-
-#cmake --build . -j ${NUM_PARALLEL_BUILDS}
+cmake --build . -j ${NUM_PARALLEL_BUILDS} --target
```
KrisThielemans, line 21: empty `--target`?

KrisThielemans (commented):
I really don't know if using `USE_SYSTEM_HDF5=OFF` is a good idea. As opposed to having 2 hdf5 builds (`apt` and `conda`), we will now have 3...

Of course, we should be able to safely set `LD_LIBRARY_PATH` to our builds (as opposed to the one from `conda`). However, presumably that means that conda installed code that uses hdf5 will break. Or maybe not, as conda's `python` presumably adds conda lib directories to its path, but then the SIRF build could break...

I think this all points to a serious problem in HDF5 shared libraries. It seems there is no foolproof way around this.

KrisThielemans:
Alternative 1:
------------
create a script (`gadgetron.sh` or even `sirf_gadgetron.sh`) which sets `LD_LIBRARY_PATH` appropriately and runs it. 

pros:
- no extra HDF5 library
- nobody else is affected.

cons:
- you have to remember to call that script
- presumably any SIRF (or STIR) executable that needs HDF5 libraries would need the same treatment, which becomes unmanageable.

Alternative 2:
-------------
build with `USE_SYSTEM_HDF5=on` but make sure gadgetron, STIR and SIRF get linked with the system version.

pros:
- no extra HDF5 library
- no need for `LD_LIBRARY_PATH`

cons:
- not sure what will happen in Python

KrisThielemans:
There's of course

Alternative 3
-------------
use `RPATH` for everything. Could work very well on `docker` as we have complete control.

docker/user_sirf-ubuntu.sh
paskino: wrong!

paskino:
CIL uses `h5py` which in turn uses `HDF5` library. These are installed by conda and get linked with `RPATH`. So, I'd argue that the conda packages will be safe. As a matter of fact, we never had an issue with CIL use of HDF5 with SIRF.

docker/user_sirf-ubuntu.sh
paskino: fixed now.

paskino:
If `USE_SYSTEM_HDF5=OFF` then the build works fine as we know where the libriaries are.
Option 3 is the best IMO. However, due to time constraints for the PSMRTBP training school, I'd like to merge this as is, possibly not closing the issue?

KrisThielemans:
Sure. Might have to be a squash merge as I guess history is very ugly here

paskino:
I always like to squash n merge

SuperBuild/External_Gadgetron.cmake
```diff
@@ -108,6 +108,7 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
       -DCMAKE_LIBRARY_PATH:PATH=${SUPERBUILD_INSTALL_DIR}/lib
       -DCMAKE_INCLUDE_PATH:PATH=${SUPERBUILD_INSTALL_DIR}/include
       -DCMAKE_INSTALL_PREFIX:PATH=${Gadgetron_INSTALL_DIR}
+      -DCMAKE_BUILD_RPATH=".;${Gadgetron_INSTALL_DIR}/lib"
```
KrisThielemans, line 111: I'm not sure what the consequences of this are, but a `.`  seems terribly dangerous

version_config.cmake
```diff
@@ -211,7 +211,7 @@ else()
 
   ## STIR
   set(DEFAULT_STIR_URL https://github.com/UCL/STIR )
-  set(DEFAULT_STIR_TAG rel_5.0.1)
+  set(DEFAULT_STIR_TAG release_5.0)
```
KrisThielemans, line 214: shouldn't be in this PR Note that 5.0.2 is out, so better to use that in a separate PR. see https://github.com/SyneRBI/SIRF-SuperBuild/pull/719

KrisThielemans (changes requested):

KrisThielemans:
I believe this is obsolete and should be abandoned

KrisThielemans:
Obsolete?

paskino:
Yes obsolete, closing

PR #718, opened 2022-05-19
docker-stacks and jupyterhub

### LINKED ISSUE #514: Create Github Actions
We need to migrate away from Travis sadly.

# caching
- general doc: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows
- `ccache`: https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/
- a precanned action based on the above: https://github.com/hendrikmuhs/ccache-action

# secrets
@casperdcl  wrote in Nov20
> Migrating saved env vars to a different provider seems the main pain point. Created this to help. Tested & works: https://github.com/casperdcl/ci-env-migrate 
>
>The advantage of github actions is that you can easily set up self-hosted runners with e.g. matlab, vis https://github.com/AMYPAD/SPM12/blob/2bd4fc4907834df0d22968dfd94d509ed16efa50/.github/workflows/test.yml#L23.

### LINKED ISSUE #563: docker building: cache misses
Buliding a new image (on my VM at least) is a painful experience as it will always reinstall conda and recompile everything in SIRF-SuperBuild. Docker uses caching to speed up building, but it's not clear to me why the caching always fails.

Here's an extract of what `./sirf-compose build sirf` writes:
```
tep 36/52 : COPY .profile .
 ---> Using cache
 ---> 8ff6a4f39ec9
Step 37/52 : RUN chmod 644 .bashrc .profile
 ---> Using cache
 ---> 5576e54f4ea0
Step 38/52 : COPY requirements.txt .
 ---> 22d78fd05aa6
Step 39/52 : COPY user_python-ubuntu.sh .
 ---> ae9c08a4382c
Step 40/52 : RUN umask 000 && bash user_python-ubuntu.sh
 ---> Running in d38e1329b314
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 17781  100 17781    0     0  17834      0 --:--:-- --:--:-- --:--:-- 17816
```
For some reason, it's step 38 that misses the cache (and hence all future steps), even though `requirements.txt` wasn't changed. Anyone any clues?

### LINKED ISSUE #592: docker: will need scripts for release images
Currently the sirf-compose-* and yml scripts do not provide for the release-service image?

### LINKED ISSUE #596: docker: remove need for 2 users and chown etc
The current set-up is fairly complicated as we have a `jovyan` user (without a lot of permissions, doing all the builds) and a `sirfuser` (with same UID:GID as host-user, such that she can read/write the mounted folders).

Other examples use `usermod` and `groupmod` to keep the same user but change UID/GID (and name?), which presumably is fast. An example is at https://github.com/sudo-bmitch/jenkins-docker/blob/main/entrypoint.sh.

I don't understand this well enough though...

### LINKED ISSUE #597: Docker - build using current SuperBuild
Per #589 , it would be logical for a user building a Docker image to have the image use the SuperBuild in its current state.

This is commonly achieved by using COPY to access the SuperBuild source in the container rather than `git pull`, but is complicated by the fact that the Dockerfile is not at the root of the directory.

Possible solution?: https://stackoverflow.com/questions/47785706/docker-using-copy-when-the-dockerfile-is-located-in-a-subdirectory

### LINKED ISSUE #652: Add GitHub Action to build jupyterhub image
With  #650 the creation of the docker image for jupyterhub has been simplified. It'd be good to add a GHA to create the new image, probably on tag/release, not on every commit to master.

### LINKED ISSUE #664: docker: to build from the local files rather than from a branch on GitHub
Currently the docker container builds only from a branch on GitHub, which makes testing a bit cumbersome as you always need to push any new thing to GitHub.

https://github.com/SyneRBI/SIRF-SuperBuild/blob/89aaac4d0b45029c30953ac954f190d9c1bdb36c/docker/user_sirf-ubuntu.sh#L6-L11

I argue that the `SIRF-SuperBuild` directory should be copied from local so that all the changes will be immediately available to test.
This may also reduce the number of `COPY` commands that we issue in the `Dockerfile` as most of the code would already have been copied in the container.

To be able to copy the whole tree one will require to issue `docker build -f docker/Dockerfile .` from the `SIRF-SuperBuild` directory.

I anticipate that this will have a number of consequences with the current docker compose files.

### LINKED ISSUE #715: jupyterhub: apt finds unsigned package repositories
Building the jupyterhub image there's an error related to unsigned repositories during `apt-get update`.

This has nothing to do with that but it is probably due to the `LD_LIBRARY_PATH` that is modified and prepends the `/opt/conda/lib` to system libraries.

See also #709

### LINKED ISSUE #763: GHA: docker run rebuilds images
Our GHA for the SIRF latest/devel looks as follows:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/844ca0f18ed334d9af91a8b18ad4390ddad81c91/.github/workflows/docker_build.yml#L101-L119
Unfortunately, at every stage, the docker image gets rebuild.:
1. The expectation would be that once `core` is built, building `sirf` wouldn't need to run all the `apt` command, as `sirf` image essentially [starts with `user_python_ubunutu/sh`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/844ca0f18ed334d9af91a8b18ad4390ddad81c91/docker/Dockerfile#L66
). All the `apt` commands get repeated though, see [here](https://github.com/SyneRBI/SIRF-SuperBuild/runs/7967521947?check_suite_focus=true#step:6:1). I guess the `docker load` statement was intended to prevent this, but it doesn't work.

2. we first build the image and then run it for the tests. But the rebuilding starts all over again, from `apt` and everything
https://github.com/SyneRBI/SIRF-SuperBuild/runs/7967521947?check_suite_focus=true#step:7:1). Weirdly, @casperdcl 's .travis.yml runs with exactly [the same line](https://github.com/SyneRBI/SIRF-SuperBuild/blob/844ca0f18ed334d9af91a8b18ad4390ddad81c91/.travis.yml#L160) and on Travis, all "core" lines in the Dockerfile [use the docker cache](https://app.travis-ci.com/github/SyneRBI/SIRF-SuperBuild/jobs/580577881#L907) as expected.

Obviously, this makes docker testing on GHA very slow/resource intensive.

@lauramurgatroyd any ideas?

### LINKED ISSUE #770: docker: service-gpu too big
The `service-gpu` image is 16 GB now (!) and it cannot be saved
https://github.com/SyneRBI/SIRF-SuperBuild/runs/8169984945?check_suite_focus=true#step:7:44
```
Error response from daemon: write /var/lib/docker/tmp/docker-export-1464324218/11af9f89cfb2e21a00a14a188b57a8e9bb64dd8001960c2b6bacfc83117a842c/layer.tar: no space left on device
```

@lauramurgatroyd I guess you struggled with this as you put in
https://github.com/SyneRBI/SIRF-SuperBuild/blob/2f6e711fbb227ffcd15e5188088cb9d897aea5a1/.github/workflows/docker_build.yml#L165-L167
However, the log says that it didn't save any disk space.

I'm not sure why the image is so big. The `service` image itself is [6.5GB](https://github.com/SyneRBI/SIRF-SuperBuild/runs/8170230815?check_suite_focus=true#step:9:33)

### LINKED ISSUE #775: GHA: docker always builds master
https://github.com/SyneRBI/SIRF-SuperBuild/blob/a0ad95c056cd7f666d8e09f296d1234227d50bef/docker/Dockerfile#L127-L128
and https://github.com/SyneRBI/SIRF-SuperBuild/blob/a0ad95c056cd7f666d8e09f296d1234227d50bef/docker/user_sirf-ubuntu.sh#L11-L13
But this means that running the docker actions on GHA doesn't make sense for most PRs (unless it actually modifies docker files). We could disable the action on these PRs I guess (with some clever tricks to make sure modifications to the `docker/*`).

At present, we're wasting a lot of computation time...

### LINKED ISSUE #804: GHA should push docker images to DockerHub
Currently we only push via Travis. This should be done via GitHub Actions.
We should follow more or less what is in `.traivs.yml` I guess. I have created the `DOCKER_PWD` and `DOCKER_USR` secrets., and did some initial testing, but abandoned it. A few things I discovered:
- the secrets are  not available in forks, so the `docker login` statement should only be used when the secrets exist.
- we should only push commits on master of course. This implies that if we make changes to `core` in a PR, just using `docker pull` (as in our `.travis.yml`) will not take advantage of caching any changes to `core`. @lauramurgatroyd 's mechanism of uploading the `core.tar.gz` artefact gets around that, but the save/upload/download itself takes probably more time than rebuilding.
- `.travis.yml` used `TRAVIS_TAG` to figure out if this was a new tag, in which case we need to create DockerHub tags as well. Not sure if this exists in GHA.

### LINKED ISSUE #810: Error when building Boost on docker
On docker, with miniconda we install whatever latest version of python miniconda comes with. I tried to fix it at 3.9 and got a build error regarding system Boost and python which are incompatible. In fact, system python on ubuntu22.04 is 3.10.

So I set `USE_SYSTEM_Boost=OFF` and I got several build errors on Gadgetron, of this type:

```
/opt/SIRF-SuperBuild/sources/Gadgetron/core/io/from_string.h:19:24: error: ‘numeric_limits’ is not a member of ‘std’
   19 |         if (tmp < std::numeric_limits<T>::min() || tmp > std::numeric_limits<T>::max())
      |                        ^~~~~~~~~~~~~~
```

### LINKED ISSUE #816: Docker: service.sh expects the cloned notebooks in the wrong directory
SIRF-Exercises are cloned in `/opt` by [default](https://github.com/SyneRBI/SIRF-SuperBuild/blob/a0d01deac13c16330327c315e0353581b4e56b5d/docker/user_service-ubuntu.sh#L4), but `service.sh` expects them in `/devel`

https://github.com/SyneRBI/SIRF-SuperBuild/blob/a0d01deac13c16330327c315e0353581b4e56b5d/docker/service.sh#L40

### LINKED ISSUE #828: docker: make sure apt cache is cleared
Make sure the apt cache is cleared before packaging the image. The following should do.

```
rm -rf /var/lib/apt/lists/*
```

paskino:
- [x] drop travis, update GHA docker CI
  + [x] push to [`ghcr.io/synerbi/sirf`](https://github.com/SyneRBI/SIRF-SuperBuild/pkgs/container/sirf)
    docker tag | SuperBuild branch/tag
    :---|:---
    `latest`, `latest-gpu` | latest tag `vM.m.p`
    `M`, `M.m`, `M.m.p`, `M-gpu`, `M.m-gpu`, `M.m.p-gpu` | tag `vM.m.p`
    `edge`, `edge-gpu` | `master`
    only build & test (no tag) | CI (current commit)
    `devel`, `devel-gpu` | `master` with `cmake -DDEVEL_BUILD=ON -DBUILD_CIL=ON`
    * [x] `build --pull` optimisation
  + [x] push to DockerHub
  + [x] `ccache` builds with https://github.com/actions/cache
- [x] auto `OMP_NUM_THREADS = cpu_count//2`
- [x] use https://github.com/jupyter/docker-stacks submodule
  + [x] update entrypoint to launch `gadgetron`
- [x] drop old docker images & scripts
  + [x] merge `/jupyterhub/` -> `/docker/`
- [x] add `docker/compose.sh` helper script
- [x] update `Dockerfile` & `docker-compose*.yml`
- [x] separate CPU & GPU images
- [x] ship CIL-Demos, CIL-Exercises, SIRF-contrib
  + [x] run `download_data.sh`
- [x] replace `make` => `ninja` builder
  + [x] auto num parallel threads
- [x] force `ipywidgets>=8` to fix widget display (at the cost of figure duplication https://github.com/TomographicImaging/CIL/issues/1600)
- [x] rebase after #862 (maybe expose more cmake options in `Dockerfile`)
- [x] update readmes
  + [x] docker permissions (`--user`, `--group-add`)
  + [ ] docker compose override permissions (`USER_ID`, `UID`)
    * `USER_ID`: you could define this as a custom override
    * `UID`: current user
    * `GID`: not needed, vis. https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations
  + [x] build options (`# CMake options` in `Dockerfile`)
  + [x] adding user deps
    ```dockerfile
    FROM synerbi/sirf:jupyter-gpu
    USER root
    RUN mamba install pytorch && fix-permissions "${CONDA_DIR}" /home/${NB_USER}
    USER ${NB_UID}
    ```
  + [x] `SIRF_DOWNLOAD_DATA_ARGS`
- [x] update changelog
- [x] use current repo state when building rather than cloning from remote URL
  + caveat: uncommitted changes are ignored for safety

----

- fixes #715
- closes #816
- part of #836
- part of #834
- fixes #828
- closes #810
- fixes #804
- fixes #775
- fixes #770
- fixes #763
- fixes #664, fixes #597
- fixes #652, fixes #514
- fixes #596
- closes #592
- fixes #563
- ~This PR is currently used to run jupyterhub for the PSMRTBP2022 school~
- [ ] However this PR changes slightly the use of `requirements.txt`
  + why?
- [ ] Probably we should use `requirements.txt` if `BUILD_CIL=OFF` or `BUILD_CIL_LITE=OFF`, and switch to requirements installed from `conda-forge` in case we build CIL.
  + why?
- [ ] Some time ago, I also made so that CIL could be installed by conda, however maintaining this will be a problem because I see that we might need to have different requirements for the case we build CIL, we want CIL but not want to build it, etc. So I propose that CIL can only be built by the SuperBuild.
  + opposite now (favour conda install)?

----

to move to follow-up issues/PRs

- [ ] add devcontainer spec
  + [ ] CPU-only default for gh codespaces
  + [ ] GPU option (e.g. using https://containers.dev/implementors/templates/#folder-structure)?


paskino:
Also I am building with the following command
```bash 
 docker build --build-arg BASE_IMAGE=nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04 \
--build-arg PYTHON_INSTALL_DIR=/opt/conda \
--build-arg EXTRA_BUILD_FLAGS="-DBUILD_CIL=ON -DIPP_LIBRARY=/opt/conda/lib -DIPP_INCLUDE=/opt/conda/include -DSTIR_URL=https://github.com/paskino/STIR.git -DSTIR_TAG=bump_parallelproj" \
--build-arg SIRF_SB_URL="https://github.com/paskino/SIRF-SuperBuild.git" \
--build-arg SIRF_SB_TAG="jupyterhub_env" --build-arg NUM_PARALLEL_BUILDS=6 --target sirf .
```

jupyterhub/Dockerfile
```diff
@@ -41,6 +41,7 @@ RUN python -m pip install torch==1.7.1 scikit-image
 # not documented in https://docs.docker.com/engine/reference/builder/#copy
 # FROM --from=nginx:latest /etc/nginx/nginx.conf /nginx.conf
 COPY --from=synerbi/sirf:sirf-core /opt/SIRF-SuperBuild/INSTALL/ /opt/SIRF-SuperBuild/INSTALL
+COPY --from=synerbi/sirf:sirf-core /opt/SIRF-SuperBuild/sources/SIRF/ /opt/SIRF-SuperBuild/sources/SIRF/
```
KrisThielemans, line 44: just curious. why?

version_config.cmake
```diff
@@ -211,7 +211,7 @@ else()
 
   ## STIR
   set(DEFAULT_STIR_URL https://github.com/UCL/STIR )
-  set(DEFAULT_STIR_TAG rel_5.0.1)
+  set(DEFAULT_STIR_TAG release_5.0)
```
KrisThielemans, line 214: this will have to change to https://github.com/UCL/STIR/commit/40ede282ad39f87faaef015c92901db25796aa6c (done that in #719 )

jupyterhub/Dockerfile
paskino: This is because we set `SIRF_PATH` that points to that directory.

version_config.cmake
paskino: sure, for the training session, however in the build command I set a number of ad hoc versions.

https://github.com/paskino/SIRF-SuperBuild/tree/jupyterhub_env/jupyterhub#start-building-sirf

jupyterhub/Dockerfile
KrisThielemans: @evgueni-ovtchinnikov I'm really not sure where `SIRF_PATH` is used. We install example data into `${CMAKE_INSTALL_PREFIX}/share/SIRF-3.2/data/` and similarly for `examples`. So I'd expect that `SIRF_PATH` should point to `${CMAKE_INSTALL_PREFIX}/share/SIRF-3.2/`. Obviously this isn't anything to fix now, but would be good to create a separate issue for this (SuperBuild I guess)

version_config.cmake
KrisThielemans: yep, you're using `origin/master` for STIR. Sadly, I've just broken that (ctest will fail), but I guess you're not running its tests...

jupyterhub/Dockerfile
paskino: Isn't it that `SIRF_PATH` is the path to SIRF's source code? It doesn't have anything to do with data, or am I mistaken?

These are the environment variables that I set in the image following [`env_sirf.sh`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/jupyterhub_env/env_sirf.sh.in)

https://github.com/paskino/SIRF-SuperBuild/blob/1d5afcbc12179fed6f9858f937363c80efb34b8e/jupyterhub/Dockerfile#L50-L55

```Dockerfile

ENV PATH "/opt/conda/bin:/opt/SIRF-SuperBuild/INSTALL/bin:$PATH"
ENV LD_LIBRARY_PATH "/opt/SIRF-SuperBuild/INSTALL/lib:/opt/SIRF-SuperBuild/INSTALL/lib64:$LD_LIBRARY_PATH"
ENV PYTHONPATH "/opt/SIRF-SuperBuild/INSTALL/python"
ENV SIRF_INSTALL_PATH "/opt/SIRF-SuperBuild/INSTALL"
ENV SIRF_EXERCISES_DATA_PATH "/mnt/materials/SIRF/Fully3D/SIRF/"
ENV SIRF_PATH "/opt/SIRF-SuperBuild/sources/SIRF"
```

version_config.cmake
KrisThielemans: fine for me. will have to update SIRF-Exercises later I guess

paskino: updated to current SuperBuild Master which points to STIR 5.1.0

jupyterhub/README.md
```diff
@@ -71,11 +78,14 @@ Please see [here](https://github.com/SyneRBI/SIRF-SuperBuild#building-ccpi-cil)
 
 ```
 
-docker build --build-arg BASE_IMAGE=nvidia/cuda:10.2-cudnn8-devel-ubuntu18.04 --build-arg PYTHON_INSTALL_DIR=/opt/conda --build-arg EXTRA_BUILD_FLAGS="-DBUILD_CIL=ON -DIPP_LIBRARY=/opt/conda/lib -DIPP_INCLUDE=/opt/conda/include -DBUILD_ASTRA=ON" --target sirf . 
-
+ docker build --build-arg BASE_IMAGE=nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04 --build-arg PYTHON_INSTALL_DIR=/opt/conda --build-arg EXTRA_BUILD_FLAGS="-DBUILD_CIL=ON -DIPP_LIBRARY=/opt/conda/lib -DIPP_INCLUDE=/opt/conda/include" --build-arg SIRF_SB_URL="https://github.com/paskino/SIRF-SuperBuild.git" --build-arg SIRF_SB_TAG="jupyterhub_env" --build-arg NUM_PARALLEL_BUILDS=6 --target sirf .
```
KrisThielemans, line 81: 18.04

jupyterhub/README.md
```diff
@@ -71,11 +78,14 @@ Please see [here](https://github.com/SyneRBI/SIRF-SuperBuild#building-ccpi-cil)
...
+```
+# build for PSMRTBP2022
+```
+ nohup docker build --build-arg BASE_IMAGE=nvidia/cuda:10.0-cudnn7-devel-ubuntu18.04 --build-arg PYTHON_INSTALL_DIR=/opt/conda --build-arg EXTRA_BUILD_FLAGS="-DBUILD_CIL=ON -DIPP_LIBRARY=/opt/conda/lib -DIPP_INCLUDE=/opt/conda/include -DSTIR_URL=https://github.com/UCL/STIR.git -DSTIR_TAG=master -DBUILD_STIR_EXECUTABLES=ON -DSIRF_URL=https://github.com/SyneRBI/SIRF.git -DSIRF_TAG=lm-recon -Dparallelproj_TAG=v0.8" --build-arg SIRF_SB_URL="https://github.com/paskino/SIRF-SuperBuild.git" --build-arg SIRF_SB_TAG="jupyterhub_env" --build-arg NUM_PARALLEL_BUILDS=6 --target sirf . > build_jupyterhub_lm.log &
```
KrisThielemans, line 85: 18.04

jupyterhub/README.md
```diff
@@ -86,7 +96,7 @@ To install SIRF we can literally _copy_ the SIRF INSTALL directory to the `datas
 
 ```
 cd SIRF-SuperBuild/docker
-docker build --build-arg BASE_IMAGE=paskino/jupyter:datascience-notebook-cuda10-cudnn8-devel-ubuntu18.04 -f ../jupyterhub/Dockerfile .
+docker build --build-arg BASE_IMAGE=paskino/jupyter:datascience-notebook-cuda10-cudnn7-devel-ubuntu18.04 -f ../jupyterhub/Dockerfile .
```
KrisThielemans, line 99: 18.04

jupyterhub/README.md
```diff
@@ -129,4 +139,4 @@ cd ../datascience-notebook
 docker build --build-arg BASE_CONTAINER=paskino/jupyter:scipy-notebook-cuda11 .
 ```
 
-Build of SIRF on Ubuntu 20.04 [fails](https://github.com/SyneRBI/SIRF-SuperBuild/issues/649)
\ No newline at end of file
+Build of SIRF on Ubuntu 20.04 [fails](https://github.com/SyneRBI/SIRF-SuperBuild/issues/649)
```
KrisThielemans, line 142: outdated

jupyterhub/README.md
```diff
@@ -51,6 +51,13 @@ docker tag 5c63287f0aee paskino/jupyter:datascience-notebook-cuda10-cudnn8-devel
 
 Finally we have the base `datascience-notebook` with the `nvidia/cuda:11.5.0-cudnn8-devel-ubuntu18.04` base image.
 
+
+Possible fix for plotting.
+```
+conda update -c conda-forge jupyterlab ipympl
+%matplotlib widget
+```
+
```
KrisThielemans, line 60: outdated

KrisThielemans (changes requested):
a few things that I spotted that might help

jupyterhub/README.md
KrisThielemans: oh sorry, it's a README. we wouldn't need these tags now. I'll resolve

KrisThielemans:
are we building ASTRA? Best for the intro notebooks.

By the way, I'm not sure that CIL dependencies are alright when building ASTRA (i.e. does it need to be guaranteed that ASTRA is built before CIL? i.e.
https://github.com/SyneRBI/SIRF-SuperBuild/blob/d50fce861da19374ddf989e4f8323b41b56f85af/SuperBuild/External_CIL.cmake#L24
might need an `if` on ASTRA

KrisThielemans:
I'm a bit surprised that all the actions are running. we're excluding `jupyterhub/**`. I thought for the docker-one could be because the white-space change in `Dockerfile`, but the others shouldn't run. oh well. Not important now, but if you don't know why this is happening, please create an issue for it after the dust settles.

paskino:
> does it need to be guaranteed that ASTRA is built before CIL?

CIL does not depend on ASTRA. We will use ASTRA if present. 

For SIRF usage it is not required. @ckolbPTB [introductory ](https://github.com/SyneRBI/SIRF-Exercises/blob/master/notebooks/Introductory/acquisition_model_mr_pet_ct.ipynb)notebook does not require ASTRA. The [gradient_descent ](https://github.com/SyneRBI/SIRF-Exercises/blob/master/notebooks/Synergistic/gradient_descent_mr_pet_ct.ipynb)one does, see https://github.com/SyneRBI/SIRF-Exercises/pull/192.

KrisThielemans:
> CIL does not depend on ASTRA. We will use ASTRA if present.

I agree, but if ASTRA is built after CIL, will it still be able to use it?

Why not build it on JupyterHub?  too late I guess.

> For SIRF usage it is not required. 

sure. https://github.com/SyneRBI/SIRF-Exercises/blob/master/notebooks/Introductory/acquisition_model_mr_pet_ct.ipynb uses it as well, although at least now it no longer fails as you've seen.

paskino:
I currently didn't add CIL to the jupyterhub image. I'll add it via conda if required. It is a 5 minute job. :D

paskino:
> if ASTRA is built after CIL, will it still be able to use it?

Yes, we only interact with ASTRA at the Python level so, as long as ASTRA and the python wrappers are built we can use it.

KrisThielemans:
> as long as ASTRA and the python wrappers are built we can use it.

yeah, but are they built if ASTRA isn't there?

KrisThielemans:
> I currently didn't add CIL to the jupyterhub image. I'll add it via conda if required. It is a 5 minute job. :D

well, if there's no CIL at all, even the intro stuff will fail, as we now just test on ASTRA.

I don't understand though. I thought you're grabbing the normal docker image, which has CIL. looks like i don't understand the jupyterhub stuff at all after all!

KrisThielemans:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fe6806f34586f0259e99ae0b639f9aa742d33d37/jupyterhub/Dockerfile#L1 is still 18.04

KrisThielemans:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fe6806f34586f0259e99ae0b639f9aa742d33d37/jupyterhub/Dockerfile#L50 says `Fully3D`. I don't care really, but just letting you know

KrisThielemans:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fe6806f34586f0259e99ae0b639f9aa742d33d37/jupyterhub/Dockerfile#L39-L40
`sirf-core`? Doesn't make sense to me. `core` doesn't have an `INSTALL`? I suppose that's your local name when building as the one on dockerhub is `sirf/core`. 

I guess it'd have to be `release-service-gpu`, except that is always going to point backwards. We currently don't seem to build `devel-service-gpu` (nor `devel-gpu`) on Travis, so we cannot pull from there.

I suppose this works for you, so no need to fix this now.

KrisThielemans:
`deprecation` isn't installed. Can't say I understand the installation of the dependencies anymore, but I'm sure @paskino does!

KrisThielemans:
`nbstripout` is not available, and there for the instructions fail. (I've updated the HackMD page)

KrisThielemans:
nibabel docopt also missing

KrisThielemans:
`parallelproj` isn't installed, so STIR doesn't have it (and neither does SIRF of course)

paskino:
> `deprecation` isn't installed. Can't say I understand the installation of the dependencies anymore, but I'm sure @paskino does!

Installed

jupyterhub/Dockerfile
```diff
@@ -1,63 +1,65 @@
-ARG BASE_IMAGE=ubuntu:18.04
+# syntax=docker/dockerfile:1
+ARG BASE_IMAGE=nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04
 # paskino/jupyter   datascience-notebook-cuda11
 FROM ${BASE_IMAGE} as base
```
casperdcl, line 4:
```suggestion
ARG BASE_CONTAINER=nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04
FROM ${BASE_IMAGE} as base
```

just using `docker-stacks` convention

jupyterhub/build_docker_stacks.sh
```diff
@@ -0,0 +1,30 @@
...
+cd ../scipy-notebook
+docker build --build-arg BASE_CONTAINER=paskino/jupyter:minimal-notebook-cuda11-cudnn8-runtime-ubuntu22.04 -t paskino/jupyter:scipy-notebook-cuda11-cudnn8-runtime-ubuntu22.04 .
+
+# datascience-notebook
+cd ../datascience-notebook
+docker build --build-arg BASE_CONTAINER=paskino/jupyter:scipy-notebook-cuda11-cudnn8-runtime-ubuntu22.04 -t paskino/jupyter:datascience-notebook-cuda11-cudnn8-runtime-ubuntu22.04 .
+
+popd
+
+docker build --build-arg BASE_IMAGE=paskino/jupyter:datascience-notebook-cuda11-cudnn8-runtime-ubuntu22.04 -t paskino/jupyter:sirf -f jupyterhub/Dockerfile .
```
casperdcl, line 30:
```suggestion

docker build --build-arg PYTHON_VERSION=3.9 --build-arg BASE_CONTAINER=nvidia/cuda:11.7.1-cudnn8-runtime-ubuntu22.04 -t synerbi/jupyter:foundation .

cd ../base-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:foundation -t synerbi/jupyter:base .

cd ../minimal-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:base -t synerbi/jupyter:minimal .


cd ../scipy-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:minimal -t synerbi/jupyter:scipy .

cd ../datascience-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:scipy -t synerbi/jupyter:datascience .

popd

docker build --build-arg BASE_CONTAINER=synerbi/jupyter:datascience -t synerbi/jupyter:sirf -f jupyterhub/Dockerfile .
```

casperdcl (changes requested):

jupyterhub/build_docker_stacks.sh
```diff
@@ -0,0 +1,24 @@
...
+
+cd ../scipy-notebook
+docker build --build-arg BASE_CONTAINER=synerbi/jupyter:minimal -t synerbi/jupyter:scipy .
+
+cd ../datascience-notebook
+docker build --build-arg BASE_CONTAINER=synerbi/jupyter:scipy -t synerbi/jupyter:datascience .
+
+popd
+
+docker build --build-arg BASE_CONTAINER=synerbi/jupyter:datascience -t synerbi/jupyter:sirf -f jupyterhub/Dockerfile .
```
casperdcl, line 24: I guess this would be safest /CC @paskino

```suggestion
pushd "$(dirname "${BASH_SOURCE[0]}")"
git submodule update --init --recursive

cd docker-stacks/images/docker-stacks-foundation
docker build --build-arg PYTHON_VERSION=3.9 --build-arg BASE_CONTAINER=nvidia/cuda:11.7.1-cudnn8-runtime-ubuntu22.04 -t synerbi/jupyter:foundation .

cd ../base-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:foundation -t synerbi/jupyter:base .

cd ../minimal-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:base -t synerbi/jupyter:minimal .


cd ../scipy-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:minimal -t synerbi/jupyter:scipy .

cd ../datascience-notebook
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:scipy -t synerbi/jupyter:datascience .

cd ..
docker build --build-arg BASE_CONTAINER=synerbi/jupyter:datascience -t synerbi/jupyter:sirf -f jupyterhub/Dockerfile .

popd
```

CHANGES.md
```diff
@@ -46,6 +46,9 @@
   - fixes documentation
   - use the VMSVGA graphics controller
   - use environment variables in Vagrantfile for easier building
+- jupyterhub updates:
+  - Ubuntu: 22.04 and nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04
```
casperdcl, line 50:
```suggestion
  - Ubuntu: 22.04 and nvidia/cuda:11.7.1-cudnn8-runtime-ubuntu22.04
```

docker/requirements.yml
```diff
@@ -30,8 +30,10 @@ dependencies:
   - ipp
   - ipp-devel
   - ipp-include
+  - dxchange
   - tigre
   - pip
+  - cil
+  - ccpi-regulariser
```
casperdcl, line 37: this would affect the old SIRF images too (not just the new docker-stacks images)... is that intended?

jupyterhub/Dockerfile
```diff
@@ -1,63 +1,64 @@
...
+
+# RUN bash ./scripts/user_service-ubuntu.sh &&\
+RUN mamba env update -n base -f docker/requirements.yml &&\
+    mamba env update -n base -f jupyterhub/requirements.yml &&\
+    git clone https://github.com/SyneRBI/SIRF-Exercises --recursive -b master SIRF-Exercises &&\
+    python scripts/install-sirf-exercises-dep.py SIRF-Exercises/requirements.txt &&\
+    mamba env update --file jupyterhub/requirements-service.yml &&\
+    mamba env update -n base -f jupyterhub/requirements.yml &&\
+    rm -rf docker && rm -rf jupyterhub &&\
+    rm -rf ./scripts
```
casperdcl, line 42: this needs some cleanup

- remove commented lines
- why is `mamba env update -n base -f jupyterhub/requirements.yml` run twice
- why are some `mamba env update` commands run with explicit `-n base` and others aren't?

jupyterhub/README.md
```diff
@@ -1,132 +1,32 @@
...
-Build the `sirf` target of the SIRF Dockerfile with the `nvidia/cuda:11.5.0-cudnn8-devel-ubuntu18.04` base image.
-
-```
-git clone git@github.com:SyneRBI/SIRF-SuperBuild.git
-cd SIRF-SuperBuild/docker
-
-# build standard SIRF docker
-docker build --build-arg BASE_IMAGE=nvidia/cuda:10.2-cudnn8-devel-ubuntu18.04 --build-arg PYTHON_INSTALL_DIR=/opt/conda --target sirf .
-
-```
```
casperdcl, line 65: I removed this section @paskino but I'm concerned it doesn't match.

The removed readme text implies you have to rebuild SIRF images using `nvidia/cuda`... but in the proposed `Dockerfile` we just `COPY --from=synerbi/sirf:latest` (i.e. the CPU SIRF image). Is this ok?

jupyterhub/README.md
```diff
@@ -1,132 +1,32 @@
...
-
-#### Building CIL
-
-Please see [here](https://github.com/SyneRBI/SIRF-SuperBuild#building-ccpi-cil) for detailed info on the command below.
-
-
-```
-
-docker build --build-arg BASE_IMAGE=nvidia/cuda:10.2-cudnn8-devel-ubuntu18.04 --build-arg PYTHON_INSTALL_DIR=/opt/conda --build-arg EXTRA_BUILD_FLAGS="-DBUILD_CIL=ON -DIPP_LIBRARY=/opt/conda/lib -DIPP_INCLUDE=/opt/conda/include -DBUILD_ASTRA=ON" --target sirf .
```
casperdcl, line 74: I also removed this section @paskino because in the proposed PR it seems CIL is just installed via `mamba`... is this correct?

casperdcl (changes requested):
just a bit concerned about which strategy is best:

1. copy SIRF installation dirs from `synerbi/sirf:latest` (CPU) image
2. copy SIRF installation dirs from `synerbi/sirf:service-gpu` (GPU) image
3. just build SIRF properly in the proposed `Dockerfile`

This PR currently does option (1), but I think (2) is safer, and (3) is best. btw option (3) also means we could perhaps deprecate the old SIRF images in favour of this new `docker-stacks` way?

casperdcl:
Also the new folder should be renamed `jupyterhub` -> `jupyter` because it has nothing to do with https://github.com/jupyterhub/jupyterhub

jupyterhub/Dockerfile
casperdcl, line 1: btw this new folder should be renamed `jupyterhub` -> `jupyter` because it has nothing to do with https://github.com/jupyterhub/jupyterhub

docker/Dockerfile
```diff
@@ -133,7 +134,7 @@ ARG NUM_PARALLEL_BUILDS="2"
 ARG BUILD_FLAGS="\
  -DCMAKE_BUILD_TYPE=Release\
  -DSTIR_ENABLE_OPENMP=ON -DUSE_SYSTEM_ACE=ON\
- -DUSE_SYSTEM_Armadillo=ON -DUSE_SYSTEM_Boost=ON\
+ -DUSE_SYSTEM_Armadillo=ON -DUSE_SYSTEM_Boost=OFF\
```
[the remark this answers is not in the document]
KrisThielemans: related issue is #810, and I think therefore you're right.

SuperBuild/External_Boost_configureboost.cmake
```diff
@@ -24,7 +24,7 @@ if(WIN32)
 else()
   message(STATUS "Build dir is : ${BUILD_DIR}")
   execute_process(COMMAND ./bootstrap.sh --prefix=${BOOST_INSTALL_DIR}
-    --with-libraries=system,filesystem,thread,program_options,chrono,date_time,atomic,timer,regex,test,coroutine,context,random
+    --with-libraries=system,filesystem,thread,program_options,chrono,date_time,atomic,timer,regex,test,coroutine,context,random,python
```
KrisThielemans, line 27: more than likely needs to be reverted. This was put in as a stop-gap, see #810

docker/requirements.yml
KrisThielemans: we came to a different conclusion in #836, but it might have been because we were in a rush

docker/user_sirf-ubuntu.sh
```diff
@@ -4,7 +4,7 @@
 # outside of the "docker build" setting
 
 [ -f .bashrc ] && . .bashrc
-set -ev
+set -v
```
KrisThielemans, line 7: probably bad idea to remove, as we do no error checking

KrisThielemans:
> just a bit concerned about which strategy is best:
> 
> 1. copy SIRF installation dirs from `synerbi/sirf:latest` (CPU) image
> 2. copy SIRF installation dirs from `synerbi/sirf:service-gpu` (GPU) image
> 3. just build SIRF properly in the proposed `Dockerfile`
> 
> This PR currently does option (1), but I think (2) is safer, and (3) is best. btw option (3) also means we could perhaps deprecate the old SIRF images in favour of this new `docker-stacks` way?

1&2 mix GPU versions, that seems weird, but in any case, I agree with you and prefer 3. The copy was done as a hack, but creates ugly dependencies.

KrisThielemans:
I can't find @casperdcl 's comment, but if docker-stacks allows merging the current `service` and `jupyterhub` images, that'd be a good thing...

docker/requirements-service.yml
```diff
@@ -7,6 +7,6 @@ channels:
 dependencies:
   - jupyterlab
   - jupyter
-  - ipywidgets
+  - ipywidgets<8
```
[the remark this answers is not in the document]
paskino: It's a [CIL requirement](https://github.com/TomographicImaging/CIL/blob/be57697c96d3e66032bac5de5246d0d276e3ba1e/recipe/meta.yaml#L98) that conda does not comply with. See also https://github.com/TomographicImaging/CIL/pull/1599

docker/requirements.yml
paskino: I don't know what would be best.

Earlier I stated https://github.com/SyneRBI/SIRF-SuperBuild/pull/718#issuecomment-1830317986 the opposite! I'm in a quantum state.

jupyterhub/Dockerfile
paskino: mistakes, I suppose.

jupyterhub/README.md
paskino: The thing is that when doing this I was building the GPU SIRF. And this doc was for me.

At the time I needed to build a specific branch/tag of SIRF(SuperBuild); these were the commands I used to build the SIRF image I required. Then I copied the files from such image to the docker-stack one.

jupyterhub/README.md
paskino: I think that having this is important for when one wants to build a particular branch/tag of CIL which is not available as conda package.

paskino:
@KrisThielemans I guess you are referring to these comments https://github.com/SyneRBI/SIRF-SuperBuild/pull/718#pullrequestreview-1754056072 and https://github.com/SyneRBI/SIRF-SuperBuild/pull/718#issuecomment-1830317986 from @casperdcl 

If I understand correctly this PR will do:

1. create the base image containing the necessary stuff to run jupyter using the official [jupyter/docker-stacks](https://github.com/jupyter/docker-stacks). Notice that they don't seem to have any release, just a master branch.
2. use the official (or built for purpose) SIRF docker image to install SIRF on the jupyter image
3. install CIL via conda dependencies that are scattered between SIRF-Exercises and SIRF-SuperBuild

I find 3 to be slighly concerning, because we need a CIL conda package if we want to install CIL now. I would like to retain the ability to build CIL via the SuperBuild and, in such case not install it via conda.

KrisThielemans:
> you are referring to these comments [#718 (review)](https://github.com/SyneRBI/SIRF-SuperBuild/pull/718#pullrequestreview-1754056072) and [#718 (comment)](https://github.com/SyneRBI/SIRF-SuperBuild/pull/718#issuecomment-1830317986) from @casperdcl

last one, yes

> I find 3 to be slighly concerning, because we need a CIL conda package if we want to install CIL now. I would like to retain the ability to build CIL via the SuperBuild and, in such case not install it via conda.

agreed, as per https://github.com/SyneRBI/SIRF-SuperBuild/issues/836#issuecomment-1739687386

Why do we "need a conda package to install CIL now"?

casperdcl:
> Why do we "need a conda package to install CIL now"?

/CC @paskino ?

> if docker-stacks allows merging the current `service` and `jupyterhub` images, that'd be a good thing...

Currently working on merging & greatly simplifying `docker` and `jupyterhub` folders, and ran into https://github.com/SyneRBI/SIRF-Exercises/pull/211 as well as #810 <- https://github.com/gadgetron/gadgetron/pull/1229... 

```sh
SIRF_SB_URL="https://github.com/SyneRBI/SIRF-SuperBuild" \
SIRF_SB_TAG="master" \
REMOVE_BUILD_FILES=1 \
RUN_CTEST=1 \
NUM_PARALLEL_BUILDS=" " \
BUILD_FLAGS="\
 -DCMAKE_BUILD_TYPE=Release\
 -DSTIR_ENABLE_OPENMP=ON -DUSE_SYSTEM_ACE=ON\
 -DUSE_SYSTEM_Armadillo=ON -DUSE_SYSTEM_Boost=??? \
 -DUSE_SYSTEM_FFTW3=ON -DUSE_SYSTEM_HDF5=OFF -DUSE_ITK=ON\
 -DUSE_SYSTEM_SWIG=ON\
 -DUSE_NiftyPET=OFF\
 -DBUILD_siemens_to_ismrmrd=ON -DBUILD_pet_rd_tools=ON" \
EXTRA_BUILD_FLAGS="-DGadgetron_USE_CUDA=ON -DBUILD_CIL=OFF -DUSE_SYSTEM_Boost=???" \
```
`bash` [`user_sirf-ubuntu.sh`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker/user_sirf-ubuntu.sh)


<details>
<summary>USE_SYSTEM_BOOST=OFF</summary>

```
...
-- Found PythonLibs: /opt/conda/lib/libpython3.9.so (found suitable version "3.9.18", minimum required is "3") 
-- Found Boost: /opt/SIRF-SuperBuild/INSTALL/lib/cmake/Boost-1.78.0/BoostConfig.cmake (found suitable version "1.78.0", minimum required is "1.65.0") found components: system program_options filesystem timer 
CMake Error at /opt/SIRF-SuperBuild/INSTALL/lib/cmake/Boost-1.78.0/BoostConfig.cmake:141 (find_package):
  Could not find a configuration file for package "boost_python" that exactly
  matches requested version "1.78.0".

  The following configuration files were considered but not accepted:

    /usr/lib/x86_64-linux-gnu/cmake/boost_python-1.74.0/boost_python-config.cmake, version: 1.74.0
    /lib/x86_64-linux-gnu/cmake/boost_python-1.74.0/boost_python-config.cmake, version: 1.74.0

  /opt/SIRF-SuperBuild/INSTALL/lib/cmake/Boost-1.78.0/BoostConfig.cmake:262 (boost_find_component)
  /opt/cmake-3.25.1-linux-x86_64/share/cmake-3.25/Modules/FindBoost.cmake:594 (find_package)
  /opt/SIRF-SuperBuild/INSTALL/lib/cmake/gadgetron/gadgetron-config.cmake:36 (find_package)
  src/xGadgetron/cGadgetron/CMakeLists.txt:44 (find_package)
```

</details>

<details>
<summary>USE_SYSTEM_BOOST=ON</summary>

```
...
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake (found suitable version "1.74.0", minimum required is "1.65.0") found components: system program_options filesystem timer
CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake:141 (find_package):
  Found package configuration file:

    /usr/lib/x86_64-linux-gnu/cmake/boost_python-1.74.0/boost_python-config.cmake

  but it set boost_python_FOUND to FALSE so package "boost_python" is
  considered to be NOT FOUND.  Reason given by package:

  No suitable build variant has been found.

  The following variants have been tried and rejected:

  * libboost_python310.so.1.74.0 (3.10, Boost_PYTHON_VERSION=3.9)

  * libboost_python310.a (3.10, Boost_PYTHON_VERSION=3.9)

  /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake:258 (boost_find_component)
  /opt/cmake-3.25.1-linux-x86_64/share/cmake-3.25/Modules/FindBoost.cmake:594 (find_package)
  /opt/SIRF-SuperBuild/INSTALL/lib/cmake/gadgetron/gadgetron-config.cmake:36 (find_package)
  src/xGadgetron/cGadgetron/CMakeLists.txt:44 (find_package)
```

</details>

any ideas?

KrisThielemans:
> Currently working on merging & greatly simplifying `docker` and `jupyterhub` folders, and ran into [SyneRBI/SIRF-Exercises#211](https://github.com/SyneRBI/SIRF-Exercises/pull/211) as well as #810 <- [gadgetron/gadgetron#1229](https://github.com/gadgetron/gadgetron/pull/1229)...

These are caused by Gadgetron-Python support. Easiest is to avoid it, for which we need https://github.com/gadgetron/gadgetron/pull/1229. This is now merged. Pretty much nothing has happened in Gadgetron except the very last commit, so I suggest trying to set  `Gadgetron_TAG=6202fb7352a14fb82817b57a97d928c988eb0f4b` (version of Oct 13, 2023).

However, *before* that merge, they merged https://github.com/gadgetron/gadgetron/pull/1225, which set ismrmrd=1.13.7, and siemens_to_ismrmrd=1.2.11, so we will need to update those versions as well (`DEFAULT_*_TAG`, not `REQUIRED`, as this is only for building them, no SIRF).

It might be best to do that in a separate PR.

Note that the `USE_SYSTEM_BOOST=OFF` case could alternatively be fixed by building Boost-Python, but as we don't Gadgetron's Python gadgets, I wouldn't go there.



> > Why do we "need a conda package to install CIL now"?
> 
> /CC @paskino ?

probably @paskino meant that currentl `BUILD_CIL=ON` is broken because of the above.

casperdcl:
i did ~a~ much thing.

~15min docker CI time (5min each for setup/build/test), or ~50min without `ccache`

Please check the updated PR description; hopefully it maketh sense @KrisThielemans @paskino :)

docker-compose.yml
```diff
@@ -0,0 +1,72 @@
...
+   args: {BASE_CONTAINER: synerbi/jupyter:base-cpu}
+ scipy:
+  command: echo nop
+  image: synerbi/jupyter:scipy-cpu
+  build:
+   context: docker/docker-stacks/images/scipy-notebook
+   args: {BASE_CONTAINER: synerbi/jupyter:minimal-cpu}
+ sirf-build:
+  volumes: [./docker/devel/.ccache:/ccache]
+  command: "bash -c 'cp -r /opt/ccache/* /ccache/'"
```
KrisThielemans, line 31: not sure why we copy? Can't we directly point ccache to /opt/ccache?

docker-compose.yml
```diff
@@ -0,0 +1,72 @@
...
+   context: .
+   target: build
+   args:
+    BASE_CONTAINER: synerbi/jupyter:scipy-cpu
+    Gadgetron_USE_CUDA: "OFF"
+    BUILD_GPU: 0
+    EXTRA_BUILD_FLAGS: >
+     -DGadgetron_TAG=6202fb7352a14fb82817b57a97d928c988eb0f4b
+     -DISMRMRD_TAG=v1.13.7
+     -Dsiemens_to_ismrmrd_TAG=v1.2.11
```
KrisThielemans, line 44: this will have to go to version_config.cmake

docker-compose.yml
```diff
@@ -0,0 +1,72 @@
...
+  build:
+   context: .
+   args:
+    BASE_CONTAINER: synerbi/jupyter:scipy-cpu
+    Gadgetron_USE_CUDA: "OFF"
+    BUILD_GPU: 0
+    EXTRA_BUILD_FLAGS: >
+     -DGadgetron_TAG=6202fb7352a14fb82817b57a97d928c988eb0f4b
+     -DISMRMRD_TAG=v1.13.7
+     -Dsiemens_to_ismrmrd_TAG=v1.2.11
```
KrisThielemans, line 58: version_config.cmake

docker-compose.yml
```diff
@@ -0,0 +1,72 @@
...
+   cache_from: [synerbi/jupyter:sirf-build-cpu]
+  stdin_open: true
+  tty: true
+  # Probably requires `export USER_ID UID` to work.
+  # Alternatively replace with the output of `id -u`.
+  user: ${USER_ID:-${UID:-${USER:-1000}}}
```
KrisThielemans, line 64: Just flagging as a concern.

I cannot quite remember all this, but we needed this for the mounting to work, but then it created trouble for file ownership/writeability (either in the container or in the host mounted dir), which we only found out when testing with a user that had a different UID/GID. So, we had to change ownership inside the container (which was difficult to do without having a large space or time penalty). Is this all sorted now?

docker/.bashrc
```diff
@@ -26,7 +26,7 @@ export PS1='sirf$ '
 
 # Need to add this as we have built using some of these shared libraries
 # See https://github.com/SyneRBI/SIRF-SuperBuild/issues/573
-export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:PYTHON_INSTALL_DIR/lib
+[ -f PYTHON_INSTALL_DIR/bin/activate ] && export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:PYTHON_INSTALL_DIR/lib
```
KrisThielemans, line 29: I don't understand this. If `activate` exists, it should be called and it should take care of `LD_LIBRARY_PATH`, no?

docker/compose.sh
```diff
@@ -0,0 +1,71 @@
...
+  c) run_cpu=1 ;;
+  g) run_gpu=1 ;;
+  r) regen_ccache=1 ;;
+  *) ;;
+  esac
+done
+# remove processed options
+shift $((OPTIND-1))
+
+test $build_cpu$build_gpu = 11 && regen_ccache=1
```
KrisThielemans, line 36: codacy will prefer lots of quotes around the expansions probably

docker/compose.sh
```diff
@@ -0,0 +1,71 @@
...
+test $build_cpu = 1 && $DCC_CPU build "$@" sirf-build
+test $build_gpu = 1 && $DCC_GPU build "$@" sirf-build
+
+echo build
+test $build_cpu = 1 && $DCC_CPU build "$@"
+test $build_gpu = 1 && $DCC_GPU build "$@"
+
+echo copy ccache
+test $regen_ccache = 1 && sudo rm -rf ./docker/devel/.ccache/*
+export USER_ID UID
```
KrisThielemans, line 63: where would it get `USER_ID` from?

`GID` not necessary?

docker/docker-compose.devel.yml
```diff
@@ -1,10 +1,23 @@
...
+version: '3.2'
 services:
+ sirf-build:
+  image: synerbi/jupyter:sirf-build-devel-cpu
+  build:
+   args:
+    EXTRA_BUILD_FLAGS: >
+     -DGadgetron_TAG=6202fb7352a14fb82817b57a97d928c988eb0f4b
+     -DISMRMRD_TAG=v1.13.7
+     -Dsiemens_to_ismrmrd_TAG=v1.2.11
```
KrisThielemans, line 10: a few more. I'll stop flagging them

docker/start-gadgetron-notebook.sh
```diff
@@ -0,0 +1,57 @@
...
+SB_PATH=$(dirname "$(dirname "$SIRF_PATH")")
+echo "start gadgetron"
+pushd "${SB_PATH}"
+test -x ./INSTALL/bin/gadgetron && ./INSTALL/bin/gadgetron >& ~/gadgetron.log&
+popd
+
+echo "make sure the SIRF-Exercises and CIL-Demos are in the expected location (~/work in the container)"
+pushd ~/work
+for repo in SIRF-Exercises CIL-Demos; do
+  test -d ${repo} || cp -dR "${SB_PATH}/../${repo}" .
```
KrisThielemans, line 40: now we have 2 copies, which is potentially confusing. Might be better to move? (but will run into user permissions probably). See also #816.

docker/start-gadgetron-notebook.sh
```diff
@@ -0,0 +1,57 @@
...
+}
+
+SB_PATH=$(dirname "$(dirname "$SIRF_PATH")")
+echo "start gadgetron"
+pushd "${SB_PATH}"
+test -x ./INSTALL/bin/gadgetron && ./INSTALL/bin/gadgetron >& ~/gadgetron.log&
+popd
+
+echo "make sure the SIRF-Exercises and CIL-Demos are in the expected location (~/work in the container)"
+pushd ~/work
```
KrisThielemans, line 38: Q: does ~/work always exist, even if not running via `docker/composesh`? (it won't be shared in the latter case)

docker/start-gadgetron-notebook.sh
```diff
@@ -0,0 +1,57 @@
...
+for repo in SIRF-Exercises CIL-Demos; do
+  test -d ${repo} || cp -dR "${SB_PATH}/../${repo}" .
+done
+./SIRF-Exercises/scripts/download_data.sh
```
KrisThielemans, line 42: I'm a bit worried about this. This script has some options to specify locations (see also https://github.com/SyneRBI/SIRF-Exercises/pull/212 for better doc).
- I'm not sure what `start-gadgetron-notebook.sh` is called, but if someone set the directories by hand to a non-default location, will they be overwritten?
- On the STFC jupterhub image, the DOWNLOAD_DIR was on a shared location for all instances, to save bandwidth/space.

docker/start-gadgetron-notebook.sh
```diff
@@ -0,0 +1,57 @@
...
+
+echo "link SIRF-Contrib into ~/work"
+if test ! -r SIRF-contrib; then
+  echo "Creating link to SIRF-contrib"
+  ln -s "${SIRF_INSTALL_PATH}/python/sirf/contrib" SIRF-contrib
```
KrisThielemans, line 47: linking here, but copying for the above. confusing.

docker/start-gadgetron-notebook.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+popd
+
+echo "start jupyter"
+start-notebook.py \
+  --PasswordIdentityProvider.hashed_password='sha1:cbf03843d2bb:8729d2fbec60cacf6485758752789cd9989e756c' \
```
KrisThielemans, line 53: not so nice to have this here as hardwired, but I suppose this is fine.

KrisThielemans (changes requested):

KrisThielemans:
Could you clarify what the proposed situation is with the file ownership and permissions?

Dockerfile
```diff
@@ -0,0 +1,129 @@
...
+FROM base as sirf
+
+# X11 forwarding
+RUN apt update -qq && apt install -yq --no-install-recommends \
+  libx11-xcb1 \
+  && apt clean \
+  && mkdir -p /usr/share/X11/xkb \
+  && test -e /usr/bin/X || ln -s /usr/bin/Xorg /usr/bin/X
+
+RUN echo "export OMP_NUM_THREADS=\$(python -c 'import multiprocessing as mc; print(mc.cpu_count() // 2)')" > /usr/local/bin/before-notebook.d/omp_num_threads.sh
```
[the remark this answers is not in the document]
casperdcl: idk; just doing what @paskino did.

NB both `nproc` and `multiprocessing.cpu_count()` yield logical processors rather than physical cores. The latter tends to be 2x less. I have no strong preference between the 3 options:

1. `//2`
2. `-1`
3. `//2 - 1`

docker-compose.yml
casperdcl: this is only a convenience function to copy `image@/opt/ccache` to `localhost@./docker/devel/.ccache`.

It's executed in `./docker/compose.sh` by `docker compose up sirf-build`.

docker-compose.yml
casperdcl: vis #851

docker-compose.yml
casperdcl: #851

docker-compose.yml
casperdcl: it's all sorted but I need to document it. TL;DR: you need to run `export UID` before running docker.

docker/.bashrc
casperdcl: prob meant to be:

```suggestion
[ -d PYTHON_INSTALL_DIR/lib ] && export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:PYTHON_INSTALL_DIR/lib
```

docker/compose.sh
casperdcl: I never put quotes around things guaranteed to have no spaces (or things where spaces are intentional, e.g. `$DCC_CPU`)

docker/compose.sh
casperdcl: - `USER_ID`: you could define this as a custom override
- `UID`: current user
- `GID`: not needed, vis. https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations

docker/start-gadgetron-notebook.sh
casperdcl: idk; this is what the old images did as well (read-only repo in `/opt` is copied to user-writable workdir on container start)

docker/start-gadgetron-notebook.sh
casperdcl: yes, it's all magically handled by the https://github.com/jupyter/docker-stacks base images

docker/start-gadgetron-notebook.sh
casperdcl: because this is meant to be read-only rather than an editable demo

docker/.bashrc
casperdcl: seems to be from https://github.com/SyneRBI/SIRF-SuperBuild/issues/573#issuecomment-862245695... maybe no longer needed>

casperdcl: removed

KrisThielemans:
#862 needs to be merged or rebased here before merge. Some work-arounds will then have to be removed.

Dockerfile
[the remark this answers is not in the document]
paskino: It's exactly to access the physical cores

paskino:
It looks good to be merged! @casperdcl @KrisThielemans

.github/workflows/docker.yml
```diff
@@ -0,0 +1,181 @@
...
+        path: docker/devel/.ccache
+    - id: build
+      name: build
+      run: |
+        suffix=${{ matrix.type == 'gpu' && '-gpu' || '' }}
+        core_image=ghcr.io/synerbi/sirf:core$suffix
+        # rebuild sirf image
+        ./docker/compose.sh -bR${{ matrix.type == 'gpu' && 'g' || 'c' }} \
+        -- --build-arg EXTRA_BUILD_FLAGS="-DDEVEL_BUILD=ON -Dsiemens_to_ismrmrd_TAG=origin/master"
+        # TODO: remove *_TAG after https://github.com/SyneRBI/SIRF-SuperBuild/issues/851
```
KrisThielemans, line 160: Obsolete comment?

Dockerfile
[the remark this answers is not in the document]
KrisThielemans: It's a wildly complicated way of getting this, but fine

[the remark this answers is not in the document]
KrisThielemans: In fact, there is no good reason to restrict to number of physical cores. Looks like you don't believe in hyperthreading, but it works ok in my experience

docker-compose.yml
KrisThielemans: This needs to be removed now as versions have been updated

docker-compose.yml
KrisThielemans: Remove as well

docker/docker-compose.devel.yml
KrisThielemans: And this

docker/requirements.yml
```diff
@@ -1,37 +1,34 @@
...
+  - dxchange         # cil
+  # - tigre          # cil (GPU)
+  # - astra-toolbox  # cil (GPU)
   - pip:
-    - git+https://github.com/data-exchange/dxchange.git  # CIL
     - git+https://github.com/ismrmrd/ismrmrd-python-tools.git@master#egg=ismrmrd-python-tools
+    # TODO: labextension @jupyter-widgets/jupyterlab-manager (jupyterlab_widgets ipywidgets ipympl)
+    # is broken because `conda-forge::cil` downgrades `ipywidgets<8` -> breaks `%matplotlib widgets`.
+    # - temp fix: install upgraded versions via pip (not conda)
+    # - proper fix: remove after https://github.com/TomographicImaging/CIL/issues/1600
```
KrisThielemans, line 31: Please create an issue for it such that we don't forget

KrisThielemans (commented):
A few final house-keeping tasks.

KrisThielemans:
@paskino I think you requested my review. My comments were entered a few days ago. you should be able to see them via "unresolved conversations". Mostly: remove specific tags (as now all in version_config), possibly increase cores. Otherwise good to go.

Dockerfile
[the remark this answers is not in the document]
casperdcl: ok you can now `docker run ... -e OMP_NUM_THREADS=99` problems:

```suggestion
RUN echo 'test -z "$OMP_NUM_THREADS" && export OMP_NUM_THREADS=$(python -c "import multiprocessing as mc; print(mc.cpu_count() // 2)")' > /usr/local/bin/before-notebook.d/omp_num_threads.sh
```

KrisThielemans (approved):
I haven't tried this, nor looked at it in detail, but overall impression is good, and my remaining worries have been resolved, so let's merge!

KrisThielemans:
surely this must close quite a few more issues than what's listed?

KrisThielemans:
Major update indeed! Let's assign it to 3.6 and merge :-)

PR #719, opened 2022-05-26
set default STIR to 40ede282 (near 5.1)

KrisThielemans:


KrisThielemans:
I think we have to close this until the listmode problems disappear. @paskino you were building with special branches anyway. if you agree, please close

KrisThielemans:
in fact, better to move 5.0.2 now.

KrisThielemans:
updated to 5.0.2 elsewhere

PR #721, opened 2022-05-29
Add Boost regex and random for Windows build

### LINKED ISSUE #720: Windows: How to specify what Boost libraries to build?
If building for Linux we specify which libraries we build for Boost
https://github.com/SyneRBI/SIRF-SuperBuild/blob/eb7d38fb8aa1d1104497d1c32fea7c68e1cbe9ac/SuperBuild/External_Boost_configureboost.cmake#L27

However, this is not available for Windows and I noticed that `random` for instance is not built, so ISMRMD fails.

paskino:
Adds `regex` and `random` to Windows Boost build, but they are not responsible for the issue #720

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # ChangeLog
 # v3.x.x
+- Windows Boost build add regex and random
 - VM: 
   - set BUILD_CIL=ON
 - updated versions:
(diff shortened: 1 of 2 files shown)


KrisThielemans:
this looks fine. but I doubt it will fix all problems. Feel free to merge it anyway

PR #735, opened 2022-07-08
use CMake 3.22 in docker scripts

KrisThielemans:
high time we upgrade CMake, but I suggest to merge this after 3.3


paskino:
Just out of curiosity, do we need 3.23?

KrisThielemans:
sure. we'll just use whatever is most recent then I guess.

KrisThielemans:
This will likely fail on docker, see #769 and https://github.com/gschramm/parallelproj/issues/17

KrisThielemans:
This PR is obselete. See #787

Issue #745, opened 2022-07-14
VM: update_VM.sh fails on GENERATE_SIMULATED_TEST_DATA 


paskino:
Once #743 is fixed, I continued the build but didn't complete. 

Got the following error

```
Scanning dependencies of target MR_PROCESS_TESTDATA
                 from /home/sirfuser/devel/buildVM/sources/SIRF/src/xGadgetron/cGadgetron/tests/mrtest_auxiliary_funs.h:32,
                 from /home/sirfuser/devel/buildVM/sources/SIRF/src/xGadgetron/cGadgetron/tests/mrtests_prep_testdata.cpp:30:
/home/sirfuser/devel/buildVM/sources/SIRF/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_image_wrap.h: In destructor ‘sirf::ImageWrap::~ImageWrap()’
:
                                                                         ^
/home/sirfuser/devel/buildVM/sources/SIRF/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_image_wrap.h:268:4: note: in expansion of macro ‘IMAGE_PROCmpl<boost::exception_detail::error_info_injector<boost::program_options::unknown_option> >'
  what():  unrecognised option '-w'
Child aborted
src/xGadgetron/cGadgetron/tests/CMakeFiles/GENERATE_SIMULATED_TESTDATA.dir/build.make:77: recipe for target 'src/xGadgetron/cGadgetron/tests/CMakeFiles/GENERATE_SIMULATED_TESTDATA' failed
```
Notice that something crashes.

paskino:
Is it a Boost error? A version problem? On the VM we are using system Boost
https://github.com/SyneRBI/SIRF-SuperBuild/blob/d9de45ba5163eb175d6bb8cd37a2031755fd4d26/VirtualBox/scripts/UPDATE.sh#L194-L198

KrisThielemans:
This is a bit confusing. The error is related to `boost::program_options::unknown_option` which seems to imply it's calling an executable that uses `program_options`. However, the lines around/above it are SIRF lines. Is this output of a parallel build?

I suspect that the SIRF lines around it are a red-herring and that it's really related to the call to `ismrmrd_generate_cartesian_shepp_logan`
[...]

PR #747, opened 2022-07-18
Fix cil astra

paskino:
Fixes updates for CIL ASTRA git-failing because an untracked file is in the sources.

Tested on updating SIRF 3.1.0 VM downloaded from [zenodo](https://zenodo.org/record/5035563#.YtRxk7fML50) with the command

```
cd devel/SyneRBI_VM
git fetch --all
git reset --hard
git checkout update_from_superbuild
# clean the SIRF-Exercises.
# if the machine has not been used it's easier to delete the SIRF-Exercises directory
cd ..
rm -rf SIRF-Exercises
.SyneRBI_VM/scripts/update_VM.sh -s -t origin/fix_cil-astra -j 1
```

This updates SIRF to 3.3.0 (master). 

Still it doesn't update CIL, for instance, but it does rebuild ISMRMRD to 1.7 with support for the utilities, thereby fixing #745

### DIFF
--- a/SuperBuild/External_CIL-ASTRA.cmake
+++ b/SuperBuild/External_CIL-ASTRA.cmake
@@ -1,6 +1,6 @@
 #========================================================================
 # Author: Edoardo Pasca
-# Copyright 2019 UKRI STFC
+# Copyright 2019-2022 UKRI STFC
 #
 # This file is part of the CCP SyneRBI (formerly PETMR) Synergistic Image Reconstruction Framework (SIRF) SuperBuild.
 #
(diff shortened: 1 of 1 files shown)


SuperBuild/External_CIL-ASTRA.cmake
```diff
@@ -50,14 +50,16 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
   if("${PYTHON_STRATEGY}" STREQUAL "PYTHONPATH")
     # in case of PYTHONPATH it is sufficient to copy the files to the
     # $PYTHONPATH directory
+    find_package(Git)
```
KrisThielemans, line 53: where needed? (no explicit  `git` commands here), for the `gitupdate`? Best to document. (Maybe we also need to have it elsewhere?)

paskino:
Failures are (around line 10000)

```
CMake Error: Error processing file: /home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/CIL-ASTRA/tmp/CIL-ASTRA-gitupdate.cmake
```

KrisThielemans (commented):
Best to document in the file why this needs to be done. We will forget...

Does  `cmake -E rm` accept multiply files to make this a bit shorter?

KrisThielemans:
do we really need to call to the `git-update`? Seems to rely quite heavily on internals of the `External` project stuff

paskino:
The rationale is:
1. I need to remove a few files from the source code otherwise git complains
2. I also want to proceed doing the standard git updates stuff that happen in `ExternalProject_Add`

As far as I understand, there is no command I can run before update. So I am hijacking it to first remove the unwanted files, then continue with the standard update that would be done in `ExternalProject_Add`

But maybe this will only work for the 3.1.0 VM?

SuperBuild/External_CIL-ASTRA.cmake
paskino: removed

KrisThielemans (approved):

PR #748, opened 2022-07-18
Github Action for Docker Builds

lauramurgatroyd:
This github action builds the docker images:
-  Starts by building (in parallel) core and core-gpu
-  Then service-gpu, devel, and latest (in parallel)
-  Then service, and devel-service (in parallel)

When each image is built, it is saved to a tar.gz file and uploaded as an artifact. (note we zip to save memory as the space for each job in a github action is limited).

Earlier builds are used to speed up the build of later builds. e.g. when devel is built, it's then uploaded as an artifact. In the devel-service job, the first thing that is done is the devel artifact is downloaded, that image is loaded, and then when the devel-service is built, it caches from the devel image. 

After all of the images have been built I have made an action which would run in parallel to load each image from the tar.gz.  We should also modify it to upload to dockerhub. I guess this would only be uploading on a tag or something like that?

Here is a diagram of the structure of the action:
image

Here is where the run of the action passed in my fork of the repo: https://github.com/lauramurgatroyd/SIRF-SuperBuild/actions/runs/2677387936 - note it takes 2 hours ish to run


KrisThielemans (changes requested):
Great. and it seems to be working!

I note we're not using any of the `sirf_compose` commands. By design I guess. It does mean that some of the smart stuff from @casperdcl isn't used. Maybe because we don't understand it...

One particular thing is the use of `ccache`. On Travis, this was pulled as an artefact, then used by the docker build, then updated back as artefact. This speeded up the whole thing rather a lot.

Yes, we do want to push this to dockerhub (`docker login` will then be essential). I have no idea if it's better to use GHA artefacts first, or immediately rely on dockerhub.

Aside from that, there's a fair amount of repetition here, but I have no idea how to reduce that. (a lot of the smart stuff in `.travis.yml` was about avoiding that).

lauramurgatroyd:
Hi Kris,

Thanks for this. Where are the sirf-compose commands used at the moment? 
I tried to use the same build/compose commands that are in the travis file.

e.g. at:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/225b7fa20bfaac05e28a7ff586be99c0d3dd164c/.travis.yml#L118-L121

Yes there is a lot of repetition. Not sure how to remove it except in some places I was able to use a matrix. I can look into what else we could do to avoid repeating so much.

KrisThielemans:
By the way, we should add exclusions in our normal workflow tests that they aren't run when only the docker workflow is modified (or the other way around). We should even exclude `Dockerfile` etc for the normal workflow. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,280 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build-core:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      
+      - name: Install docker and pre-reqs
```
casperdcl, line 13: consider https://github.com/docker/build-push-action#usage to simplify a lot of this

casperdcl (commented):
good effort, I don't really have time to review.

I do like the parallelised builds by (ab)using art[ie]fact dependencies though :)

.github/workflows/docker_build.yml
lauramurgatroyd: Thanks, unfortunately I don't have much more time I can spend on this, so can't really look into this at the moment. It's something we could do in the future though.

I had briefly looked at using that action originally, but I couldn't see how to separate out the build and publish steps, which I thought may be worthwhile keeping because if there's a network error you may want to just rerun the publish step without wasting time to rerun the build

paskino:
Isn't ccache used by the docker image?
https://github.com/lauramurgatroyd/SIRF-SuperBuild/blob/6d95a53771ead7aba1177d0b71fc1c0ce50fd930/docker/build_essential-ubuntu.sh#L31-L35

Can it be used between docker containers? That'd be quite a nice feature, I can't think how it could be actually done...

KrisThielemans:
@casperdcl  set `ccache` up for [Travis here](https://github.com/SyneRBI/SIRF-SuperBuild/blob/3ba4dc86ff6c8ee7ffab76369b68a5f394809639/.travis.yml#L135-L151).  I believe it essentially does this:
- let Travis/GHA manage the `~/.ccache` folder as normal
- in the CI script, copy the `..ccache`  folder to the container
- run the container as normal
- at the end, copy the container's cache to `~/.cache`

Sounds elegant but actual lines are bit long. We certainly wouldn't want to duplicate them many times. Maybe some of it could be put in its own script.

KrisThielemans:
@lauramurgatroyd can you put some excludes in the yml files such that e.g. "normal" jobs don't get run when only the docker yml (and even `Dockerfile`) gets changed? And the other way around.

At some point, it'd be nice to also not run the "release" docker images until it's really doing something different (not sure if that's possible/desirable).

lauramurgatroyd:
We could make the following updates, using this action to help: https://github.com/marketplace/actions/changed-files 

1. https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/.github/workflows/c-cpp.yml does not get run if the only files changed are in this folder: https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker or the docker action files.

2. The docker action does not get run if the only file changed is: https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/.github/workflows/c-cpp.yml . My understanding is that with any other changes the docker action would want to run so that it is running the tests?

Is the above correct? Are there any other conditions that should be set?

Please remind me which are the "release" docker images and when they should be run?

paskino:
We cannot exclude all of the `docker` directory, because all actions make use of files in it. But we may exclude the `Dockerfile` file as the `c-cpp.yml` action doesn't use it.

lauramurgatroyd:
Oops, I forgot!

KrisThielemans:
I think `changed-files` is too complicated for what we need. `paths-ignore` should do the trick (and is far easier to understand). see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore. I guess stuff like this in our "main` action.
```
on:
  push:
    branches: [ master ]
    paths-ignore:
     - '.github/workflows/*docker*'
      - 'docker/Dockerfile'
      - 'docker/*yml`
      - '**.md'
```
and similar for the docker action.
(I don't know how to avoid duplication between the branches and pull-requests entries).

KrisThielemans:
> Please remind me which are the "release" docker images and when they should be run?

"release" is "latest" in the docker terminology. I feel they should only be run on a tagged version, but  I might be wrong.

lauramurgatroyd:
Hi Kris, I added "path-ignore"s to both of the actions. But note, on this PR it is still running the c-cpp.yml action because of the fact I had to change that file to add the paths to ignore.

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,291 @@
...
+        shell: bash -l {0}
+        run: |
+         # make the file runnable
+         chmod +x ".github/workflows/install_docker.sh"
+         .github/workflows/install_docker.sh
+         
+      - name: setup variables
+        shell: bash -l {0}
+        run: |
+          export compose_command="docker-compose -f docker/docker-compose.yml"
```
KrisThielemans, line 203:
```suggestion
          compose_command="docker-compose -f docker/docker-compose.yml"
```

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,291 @@
...
+           
+          if [[ ${{ matrix.tag }} == 'service' ]]; 
+          then
+            echo "prereq_image_name="latest"" >> $GITHUB_ENV
+            export compose_command="$compose_command -f docker/docker-compose.srv.yml"
```
KrisThielemans, line 208:
```suggestion
            compose_command="$compose_command -f docker/docker-compose.srv.yml"
```

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,291 @@
...
+          fi
+          if [[ ${{ matrix.tag }} == 'devel-service' ]]
+          then
+            echo "prereq_image_name="devel"" >> $GITHUB_ENV
+            export compose_command="$compose_command -f docker/docker-compose.devel.yml -f docker/docker-compose.srv.yml"
```
KrisThielemans, line 213:
```suggestion
            compose_command="$compose_command -f docker/docker-compose.devel.yml -f docker/docker-compose.srv.yml"
```

KrisThielemans (changes requested):
remove a few more `export`. I'll commit this

.github/workflows/docker_build.yml
```diff
@@ -0,0 +1,280 @@
...
+          # $compose_command pull core
+          $compose_command build --pull sirf
+          echo "compose_command=$compose_command" >> $GITHUB_ENV
+
+      - name: Run docker container with tests
+        shell: bash -l {0}
+        run: |
+          ${{ env.compose_command }} run --rm -u jovyan --entrypoint /bin/bash sirf --login -c /devel/test.sh 1
+          
+      - name : docker save image
```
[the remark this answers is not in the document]
KrisThielemans: @lauramurgatroyd I think this just means swapping the test and save targets around. Want to give that a go?

KrisThielemans (approved):
ok. This all worked. Scope for improvement, but I'll merge this now.

Issue #750, opened 2022-08-01
create SIRF_DISABLE_Gadgetron etc


KrisThielemans:
SIRF has CMake variables `DISABLE_Gadgetron` etc. Currently these are not exposed in `External_SIRF.cmake`. That means that they can only be set via `SIRF_EXTRA_CMAKE_ARGS="-DDISABLE_Gadgetron:BOOL=OFF"`. (Similar variables **are** exposed in `External_STIR.cmake`). 

To make matters confusing, there are `BUILD_*` variables in the Superbuild. There is some logic related to the `BUILD_Gadgetron` etc variables
https://github.com/SyneRBI/SIRF-SuperBuild/blob/02d30f86dccbfe7d0de1c2bd55a8a6e0deb9c5ee/SuperBuild/External_SIRF.cmake#L28-L39
This says that if you set `BUILD_Gadgetron=OFF`, Gadgetron is not added as a dependency to the SIRF project (the `ExternalProject` dependencies ensure build order). That is fine in itself, but it does **NOT** mean that the SIRF build will not look for ISMRMRD (and optionally Gadgetron toolboxes).

Note that on Windows, we set `BUILD_Gadgetron=OFF` such that we don't attempt to build Gadgetron, but Gadgetron support in SIRF is still enabled (by intention).

## Conclusion

if you don't need MR support in SIRF, at present you need to
```
cmake -DBUILD_Gadgetron:BOOL=OFF -DSIRF_EXTRA_CMAKE_ARGS="-DDISABLE_Gadgetron:BOOL=OFF" ....
```
Not great...

## More on naming ##

In addition, there are the `USE_SYSTEM_*`  variables. You can set `BUILD_Gadgetron=ON` but `USE_SYSTEM_Gadgetron=ON`. which means that Gadgetron wll not be built. Confusing...

See some more discussion on naming in #511.

PR #751, opened 2022-08-02
pass DISABLE_Gadgetron to SIRF

paskino:
Partly addresses #750 by setting `DISABLE_Gadgetron=(NOT ${BUILD_Gadgetron}) `in SIRF's external build.

This is however, still not entirely correct as it does not address:

1. `USE_SYSTEM_Gadgetron=ON`
2. the case in which `BUILD_Gadgetron=ON` but the user wants (why would they?) to disable Gadgetron's support in SIRF.


KrisThielemans:
I'm not really sure if `NOT` works that way. Does it?

In any case, I don't think this will cope with what we want on Windows.  There `BUILD_Gadgetron=OFF` but (at least normally) SIRF's `DISABLE_Gadgetron=OFF`.

I see no other way than introducing `SIRF_DISABLE_Gadgetron`.

paskino:
It does work, as I used it to build on my Windows machine lately. However, it doesn't fullfil the requirements and `SIRF_DISABLE_Gadgetron` is a simpler solution.

PR #752, opened 2022-08-02
trims to minimum requirements_conda_forge

paskino:
The idea here is that CIL will be only built within the SuperBuild and cannot be installed with the conda-forge requirements


KrisThielemans:
presumably this is to fix
https://app.travis-ci.com/github/SyneRBI/SIRF-SuperBuild/jobs/578510836#L2588-L2589

One more piece of evidence that mixing Python package managers is a bad idea. A pity that none of them supplies all we need...

Good lucking sorting out this mess!

paskino:
Yes! Also I don't understand why `llvmlite` can't be updated while `numpy` and other can ...

Anyway, here I am trying to install everything via pip except the IPP which is available via intel's conda channel, and tigre which is available from the ccpi channel.

KrisThielemans:
Once we know it works, I suggest to create a new PR without the docker GHA and merge that.

KrisThielemans:
well ok, maybe after attempting Travis as well

paskino:
The gha worked! Travis needs triggering

KrisThielemans:
sorry. doesn't work. I first pushed it SyneRBI as opposed to your fork. But Travis was configured to only run `master`. So I tried to allow allow branches. still doesn't work. As life is short, I vote for merging only the relevant commit to master. Then we can trigger travis.

paskino:
superseded by #753

PR #786, opened 2022-12-13
GHA updates

KrisThielemans:


KrisThielemans:
the ccache size has been taken, we have a lot of cache hits, but still only need 309MB, so maybe this increase wasn't necessary. anyway... doesn't harm I guess.

PR #438, opened 2020-10-06
Build gadgetron master

paskino:
This changes should allow the build of Gadgetron master.
closes #378 
closes #384

PR #787, opened 2023-01-05
Release 3.4

paskino:
Final tweaks for release 3.4 of SIRF & Co.

With this PR we aim at updated versions:

- [x] ubuntu 22.04
- [x] SIRF latest commit/3.4 tag
- [x] STIR rel_5.1.0
- [x] parallelproj 1.2.13
- [x] Boost 1.78
- [ ] Gadgetron latest commit


paskino:
errors are due to the missing [`pytest-cov`](https://github.com/pytest-dev/pytest-cov) package

docker/requirements.yml
```diff
@@ -0,0 +1,35 @@
...
+  - wget         # CIL
+  - six          # CIL
+  - olefile      # CIL
+  - opencv       # CIL
+  - numba        # CIL
+  - pandas
+  - tifffile
+  - nibabel
+  - deprecation
+  - nose
```
KrisThielemans, line 28: shouldn't need `nose` anymore

docker/requirements.yml
```diff
@@ -0,0 +1,35 @@
...
+  - ipp
+  - ipp-devel
+  - ipp-include
+  - tigre
+  - pip:
+    - git+https://github.com/data-exchange/dxchange.git  # CIL
+    - git+https://github.com/ismrmrd/ismrmrd-python-tools.git@master#egg=ismrmrd-python-tools
```
KrisThielemans, line 35: use https://anaconda.org/inati/ismrmrd-python-tools presumably

KrisThielemans (changes requested):
we had `pytest-cov` in `requirements.txt` already, so you might want to check about any other changes there that didn't make it into your `yml`

docker/requirements.yml
paskino: It says "Last upload: 7 years and 3 months ago" I wouldn't use it! Also it is version 0.10.0 for osx.

KrisThielemans: oh well.

@DANAJK @ckolbPTB @evgueni-ovtchinnikov should we still be installing ismrmrd-python-tools? (its github is more up-to-date: Oct 2020(

paskino:
I moved to Ubuntu 22.04 and stumbled upon some errors with CUDA.

```
nvcc fatal   : Unsupported gpu architecture 'compute_86+PTX'
```

KrisThielemans:
building which project?

paskino:
> building which project?

`parallelproj`

KrisThielemans:
Possible @gschramm can help, but a quick look at the internet says this is a conflict between driver and toolkit version

paskino:
using parallelproj v1.2.11 solved the issue, now the next one is

```
Linking CXX executable ismrmrd_test_xml
ESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_global_init@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_easy_perform@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_slist_free_all@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_easy_setopt@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_easy_init@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_slist_append@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_easy_cleanup@CURL_OPENSSL_4'
ESC[0mESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_global_cleanup@CURL_OPENSSL_4'
```

KrisThielemans:
at least one job is failing because `parallelproj` by defaults requires `nvcc`, see e.g. https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3849192276/jobs/6557897239#step:7:507 and https://github.com/gschramm/parallelproj/issues/22. If no CUDA, we need to build it with `SKIP_CUDA_LIB`. Easy?

KrisThielemans:
```
> ESC[91m/usr/bin/ld: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so: undefined reference to `curl_global_init@CURL_OPENSSL_4'
```
I guess this is on a local build as I don't see it in any of the logs. What configuration is this? (As a 22.04 VM builds, and @ckolbPTB builds (of course it's [the SB `master` version](https://github.com/ckolbPTB/OpenSourceMrRecon/blob/1927ac48227454d3f3d57ec382632feb706693df/sirf_gadgetron.sh#L28) on docker 22.04 https://github.com/ckolbPTB/OpenSourceMrRecon/blob/main/Dockerfile)

This might be an Ubuntu (possibly it distributes built curl without openssl) or CMake problem (possibly it doesn't know what libraries it should pass).

docker/requirements.yml
evgueni-ovtchinnikov: We use `ismrmrd-python-tools` only for the coil sensitivity computation with Inati method (an alternative to our default Square-Root-of-the-Sum-of-Squares method): 
```
Searching 'ismrmrd-python-tools' in C:\Users\wps46139\Documents\GitHub\build\sources\SIRF\*.py ...
C:\Users\wps46139\Documents\GitHub\build\sources\SIRF\examples\Python\MR\coil_sensitivity_maps.py: 116:         print('Inati method requires ismrmrd-python-tools')
C:\Users\wps46139\Documents\GitHub\build\sources\SIRF\src\xGadgetron\pGadgetron\Gadgetron.py: 629:                 raise error('Inati method requires ismrmrd-python-tools')
C:\Users\wps46139\Documents\GitHub\build\sources\SIRF\src\xGadgetron\pGadgetron\Gadgetron.py: 655:                 raise error('Inati method requires ismrmrd-python-tools')
Hits found: 3
```

paskino:
Yes it is a local build. 

This is a Ubuntu22.04 from a 11.7.1-cudnn8-devel-ubuntu22.04 [NVIDIA image](https://hub.docker.com/layers/nvidia/cuda/11.7.1-cudnn8-devel-ubuntu22.04/images/sha256-60a84ed05ef265b02ba03a3c8172fed160b3a0ec1d5d4c982b7f4f21d688b07e?context=explore). 

The build command is 

```
docker build --build-arg BASE_IMAGE=nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04 \
  --build-arg PYTHON_INSTALL_DIR=/opt/conda --build-arg SIRF_SB_URL="https://github.com/paskino/SIRF-SuperBuild" \
  --build-arg SIRF_SB_TAG="fix_prerelease3.4" --build-arg NUM_PARALLEL_BUILDS=1 --target sirf .
```

With `USE_SYSTEM_HDF5=ON` I get the linking error.

I will try to install `libcurl4-openssl-dev` although I thought I checked and verified it was already installed.

KrisThielemans:
Probably unrelated, but in STIR GHA tests, I suddenly [get](https://github.com/UCL/STIR/actions/runs/3863614141/jobs/6585904912#step:4:153)
```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/c/curl/libcurl4-openssl-dev_7.81.0-1ubuntu1.6_amd64.deb  404  Not Found [IP: 40.119.46.219 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```
This library is installed because of HDF5. So maybe they're fixing something. (I had to `apt update` first)

paskino:
Yes it was installed

```
Step 62/76 : RUN apt-get install -y libcurl4-openssl-dev
 ---> Running in 5c8a5c97be87
Reading package lists...
Building dependency tree...
Reading state information...
libcurl4-openssl-dev is already the newest version (7.81.0-1ubuntu1.6).
libcurl4-openssl-dev set to manually installed.
0 upgraded, 0 newly installed, 0 to remove and 7 not upgraded.
```

However, on the ubuntu [page ](https://packages.ubuntu.com/jammy/libcurl4-openssl-dev) the version is 7.81.0-1ubuntu1.7 and it is a security update. I guess we need to run an `apt upgrade` before cmaking.

paskino:
Ah, I understand why I had the `7.81.0-1ubuntu1.6` version instead of `7.81.0-1ubuntu1.7` of `libcurl4-openssl-dev`! 

This is because I built using cached images. So my image, although it had run `apt-get update`, it had done it when the current version was 1.6. Let's hope this fixes the problem.

paskino:
It seems that we should link ISMRMRD adding `-lcurl` and it works.

KrisThielemans:
> It seems that we should ~link~ build ISMRMRD adding `-lcurl` and it works.

Only necessary for ISMRMRD and not for SIRF? Weird.

I doubt they use curl directly, so think this is a CMake bug. But I cannot understand why you see this in your docker image and we don't have it elsewhere. Which version of CMake are you using in your docker instance?

> How can I do this?

We could force the linker flag `CMAKE_EXE_LINKER_FLAGS` (ugly), for `HDF5_LIBRARIES` variable (ugly) or use our own clone for ISMRMRD (not nice). But as they don't have this in CI, nor we as far as I understand, I'd want to understand better why you have it...

paskino:
The problem seems to arise when using the system HDF5 library, not if we build our own HDF5.
I just fixed it for ISMRMRD and it appeared for Gadgetron.

My CMake is v3.25.1

KrisThielemans:
> The problem seems to arise when using the system HDF5 library, not if we build our own HDF5. I just fixed it for ISMRMRD and it appeared for Gadgetron.
> 
> My CMake is v3.25.1

strange. definitely a CMake bug then. Can you post some log when building ISMRMRD after `find_package(HDF5)` (I think we have the `HDF5_FIND_DEBUG` flag on such that it writes diagnostics, but if not, please run with it

paskino:
```
-- SuperBuild -     HDF5[OK]
-- HDF5 C compiler wrapper is unable to compile a minimal HDF5 program.
-- Found HDF5: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so (found suitable version "1.10.7", minimum required is "1.8")
-- USING the system HDF5, found HDF5_INCLUDE_DIRS=/usr/include/hdf5/serial, HDF5_C_LIBRARY_hdf5=,HDF5_LIBRARIES=/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so
-- HDF5_CMAKE_ARGS=-DHDF5_INCLUDE_DIRS:PATH=/usr/include/hdf5/serial;-DHDF5_LIBRARIES:STRING=/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so;-DHDF5_C_LIBRARIES:STRING=;-DHDF5_FIND_DEBUG:BOOL=ON
```

For ISMRMRD

```
[ 18%] Performing configure step for 'ISMRMRD'
loading initial cache file /opt/SIRF-SuperBuild/builds/ISMRMRD/tmp/ISMRMRD-cache-Release.cmake
-- The C compiler identification is GNU 11.3.0
-- The CXX compiler identification is GNU 11.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib/ccache/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib/ccache/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- HDF5 C compiler wrapper is unable to compile a minimal HDF5 program.
-- Found HDF5: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so;/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so (found version "1.10.7") found components: C
-- HDF5_DIR: HDF5_DIR-NOTFOUND
-- HDF5_DEFINITIONS:
-- HDF5_INCLUDE_DIRS: /usr/include/hdf5/serial
-- HDF5_LIBRARIES: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so;/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so
-- HDF5_HL_LIBRARIES:
-- HDF5_C_DEFINITIONS:
-- HDF5_C_INCLUDE_DIR: /usr/include/hdf5/serial
-- HDF5_C_INCLUDE_DIRS: /usr/include/hdf5/serial
-- HDF5_C_LIBRARY:
-- HDF5_C_LIBRARIES: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so
-- HDF5_C_HL_LIBRARY:
-- HDF5_C_HL_LIBRARIES:
-- Defined targets (if any):
-- ... hdf5::hdf5
mHDF5 found at: /usr/include/hdf5/serial
HDF5 found at: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so
-- Found Git: /usr/bin/git (found version "2.34.1")
-- Found Doxygen: /usr/bin/doxygen (found version "1.9.1") found components: doxygen dot
   Building info application
-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake (found suitable version "1.74.0", minimum required is "1.43") found components: program_options random
-- FFTW3 UNIX libraries: /usr/lib/x86_64-linux-gnu/libfftw3.so/usr/lib/x86_64-linux-gnu/libfftw3f.so
   FFTW3 and Boost Found... building utilities
-- Found FFTW3: /usr/lib/x86_64-linux-gnu/libfftw3f.so
CMake Deprecation Warning at examples/c/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.74.0/BoostConfig.cmake (found suitable version "1.74.0", minimum required is "1.43") found components: unit_test_framework
   CPACK_PACKAGING_INSTALL_PREFIX: /opt/SIRF-SuperBuild/INSTALL
-- Found CPack generators: DEB
-- Configuring done
-- Generating done
```

paskino:
Obviously adding `-lcurl` doesn't work on Ubuntu < 22.04 https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3874884037/jobs/6606658793#step:7:3764

KrisThielemans:
Please upgrade boost (like in #438). 1.72 gives problems with ISMRMD.

KrisThielemans:
The `nvcc fatal   : Unsupported gpu architecture 'compute_86+PTX'` issue is not related to parallelproj version. We tried on an Ubuntu 22.04 system with parallelproj 1.2.12. I still think this is a conflict between nvcc versions etc.

Currently we're trying with CUDA toolkit 12.0 from the NVIDIA website. Note that this says to use `apt install cuda`, using `apt install nvidia-cuda-toolkit` then creates a conflict as it installs nvcc 11.5 back again. Will let you know.

KrisThielemans:
> Currently we're trying with CUDA toolkit 12.0 from the NVIDIA website. Note that this says to use `apt install cuda`, using `apt install nvidia-cuda-toolkit` then creates a conflict as it installs nvcc 11.5 back again. Will let you know.

`apt install cuda` didn't install `nvcc`, and it says you need `apt install nvidia-cuda-toolkit` anyway. But that downgraded. Oh well.

paskino:
Wait @KrisThielemans ! The problem is that the the docker build was using a Ubuntu 20.04 image as base. I changed that and it builds.  Currently it fails installing jupyter, but I'm probably close to fix it. 

I will need to upgrade Boost.

paskino:
Next problem is installing SIRF-Exercises dependencies when we use conda...

paskino:
An attempt to build Gadgetron master is https://github.com/paskino/SIRF-SuperBuild/pull/4/

KrisThielemans:
> Next problem is installing SIRF-Exercises dependencies when we use conda...

we never had a problem with that in the past, so do we have one now? I guess `pip install` should still work?

KrisThielemans:
@paskino what's the situation with HDF5 now? I guess you're building your own? I've checked in my 22.04 VM and spotted the difference that explains why it works on the VM and not for you. On my 
VM, the CMake output is
```
irfuser@vagrant:~/devel/buildVM/builds/ISMRMRD/build$ cmake .
-- HDF5: Using hdf5 compiler wrapper to determine C configuration
-- HDF5_DIR: HDF5_DIR-NOTFOUND
-- HDF5_DEFINITIONS: 
-- HDF5_INCLUDE_DIRS: /usr/include/hdf5/serial
-- HDF5_LIBRARIES: /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so;/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so;/usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.so;/usr/lib/x86_64-linux-gnu/libcrypto.so;/usr/lib/x86_64-linux-gnu/libcurl.so;/usr/lib/x86_64-linux-gnu/libpthread.a;/usr/lib/x86_64-linux-gnu/libsz.so;/home/sirfuser/devel/install/lib/libz.a;/usr/lib/x86_64-linux-gnu/libdl.a;/usr/lib/x86_64-linux-gnu/libm.so
...
```
The first line says that it uses `h5cc` to find the options, while yours says it couldn't use the compiler wrapper. Indeed, on my VM I have
```
$ h5cc -show
gcc -I/usr/include/hdf5/serial -L/usr/lib/x86_64-linux-gnu/hdf5/serial /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5_hl.a /usr/lib/x86_64-linux-gnu/hdf5/serial/libhdf5.a -lcrypto -lcurl -lpthread -lsz -lz -ldl -lm -Wl,-rpath -Wl,/usr/lib/x86_64-linux-gnu/hdf5/serial
```
Checking https://packages.ubuntu.com, it appears that `h5cc` sits in `hdf5-helpers`. So, I think that if you'd add that to our last of apt packages, you can switch back to using the system libraries.

paskino:
> > Next problem is installing SIRF-Exercises dependencies when we use conda...
> 
> we never had a problem with that in the past, so do we have one now? I guess `pip install` should still work?

`pip install` works but wants to uninstall `llmvlite` installed by conda and refuses to do it, so it breaks.

KrisThielemans:
ah well. Probably a few ways around it:
- use `conda install requirements.txt` as opposed to `pip install requirements.txt`
- add a specific `requirements.yml` for `conda` in the SIRF-exercises
- pin version numbers (I guess python people would recommend that, but I generally try to avoid it as these things do get out-dated).

Maybe you can show a more complete error log?

docker/user_python-ubuntu.sh
```diff
@@ -52,14 +52,17 @@ miniconda)
 esac
 
 if [ "$PYTHON" = "miniconda" ]; then
-  if [ -f requirements_conda_forge.txt ]; then
-    conda install --yes -c conda-forge -c intel -c ccpi -c astra-toolbox --file requirements_conda_forge.txt
-  fi
+  # updates the various packages
   conda update -c conda-forge -y --all
```
KrisThielemans, line 56: not sure if this update is a great idea. It does generally take ages, but it might then be changed by the next step anyway. Could leave it in for now.

docker/user_service-ubuntu.sh
```diff
@@ -2,21 +2,38 @@
...
+  fi
+  conda clean -y --all
+# Python (runtime)
+else
+  if [ -f requirements-service.txt ]; then
+    conda install -c conda-forge -y --file requirements-service.txt || \
+    pip install -U -r requirements-service.txt
+  fi
+  #install SIRF-Exercises requirements
+  cd $INSTALL_DIR/SIRF-Exercises
```
KrisThielemans, line 30: This `cd` is now inside the `if`. That means it isn't executed when using `miniconda`, which then leads to an error in `nbstripout --install` as we're not in the correct place.

docker/user_service-ubuntu.sh
```diff
@@ -2,21 +2,38 @@
...
+  if [ -f requirements-service.yml ]; then
+    # installs the required packages in the environment with requirements.yml. 
+    # Notice that SciPy is set to 1.7.3 to prevent `GLIBCXX_3.4.30' not found
+    conda env update --file requirements-service.yml 
+  fi
+  conda clean -y --all
+# Python (runtime)
+else
+  if [ -f requirements-service.txt ]; then
+    conda install -c conda-forge -y --file requirements-service.txt || \
```
KrisThielemans, line 26: It makes no sense to have a `conda install` in the non-conda case

docker/user_service-ubuntu.sh
```diff
@@ -2,21 +2,38 @@
...
+    pip install -U -r requirements-service.txt
+  fi
+  #install SIRF-Exercises requirements
+  cd $INSTALL_DIR/SIRF-Exercises
+  if [ -f requirements.txt ]; then
+    conda install -c conda-forge -y --file requirements.txt || \
```
KrisThielemans, line 32: no conda here

KrisThielemans (changes requested):
some comments on the python stuff. In addition, at the moment, when using miniconda, we don't install `SIRF-Exercises/requirements.txt`, but I guess that is by intention as you mentioned the `llvmlite` problem.

docker/requirements.yml
KrisThielemans: I suppose we leave this in then, as it doesn't seem to cause problems.

KrisThielemans:
@paskino, I suggest that we merge this PR after fixing the following:
- Boost version
- `hdf5-helpers`
- your current SIRF-Exercises problem

i.e. leave Gadgetron master for a separate PR (which you have already anyway).

This PR will get us to a working version on Ubuntu 22.04, which more and more people are trying of course.

docker/user_service-ubuntu.sh
paskino: it should be fixed now

paskino:
> ah well. Probably a few ways around it:
> 
> * use `conda install requirements.txt` as opposed to `pip install requirements.txt`
> * add a specific `requirements.yml` for `conda` in the SIRF-exercises
> * pin version numbers (I guess python people would recommend that, but I generally try to avoid it as these things do get out-dated).
> 
> Maybe you can show a more complete error log?

The error log is [here](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3892212789/jobs/6645282809#step:7:2249)

The actual error is `CondaValueError: could not parse 'numpy --only-binary=numpy' in: requirements.txt` 
So, conda install of the requirements of SIRF-Exercises fails because of that `--only-binary=numpy` and then everything goes to `pip`. Maybe the solution is to remove the `--only-binary` stuff from https://github.com/SyneRBI/SIRF-Exercises/blob/master/requirements.txt

paskino:
My current solution works, but [the failure](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3900437742/jobs/6661448605) is due to a `no space left of device` problem on the GPU runner. Apparently the build of the `sirf` target and the `service` one are not separated in the GPU case.

SuperBuild/External_ISMRMRD.cmake
```diff
@@ -65,6 +65,7 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
       -DCMAKE_INSTALL_PREFIX:PATH=${ISMRMRD_INSTALL_DIR}
       -DCMAKE_PREFIX_PATH:PATH=${SUPERBUILD_INSTALL_DIR}
       -DCMAKE_LIBRARY_PATH:PATH=${SUPERBUILD_INSTALL_DIR}/lib
+      # -DCMAKE_EXE_LINKER_FLAGS:STRING="-lcurl"
```
KrisThielemans, line 68: let's remove this again

docker/build_system-ubuntu.sh
```diff
@@ -56,6 +56,7 @@ fi
 ${APT_GET_INSTALL} \
   libfftw3-dev         \
   libhdf5-serial-dev   \
+  hdf5-helpers         \
```
KrisThielemans, line 59: best to add a comment here why we needed it (i.e. for CMake discovery of libraries needed for HDF5)

docker/user_service-ubuntu.sh
```diff
@@ -2,27 +2,47 @@
...
 
-#install SIRF-Exercises requirements
-cd $INSTALL_DIR/SIRF-Exercises
-if [ -f requirements.txt ]; then
-  conda install -c conda-forge -y --file requirements.txt || \
-  pip install -U -r requirements.txt
+# Python (runtime)
+else
+  if [ -f requirements-service.txt ]; then
+    pip install -U -r requirements-service.txt
```
KrisThielemans, line 27: maybe this is running in a venv, but I guess this might be safer
```suggestion
    $PYTHON_EXECUTABLE -m pip install -U -r requirements-service.txt
```

docker/user_service-ubuntu.sh
```diff
@@ -2,27 +2,47 @@
...
+  fi
+  #install SIRF-Exercises requirements
+  cd $INSTALL_DIR/SIRF-Exercises
+  if [ -f requirements.txt ]; then
+    python ~/install-sirf-exercises-dep.py requirements.txt
```
KrisThielemans, line 32:
```suggestion
    $PYTHON_EXECUTABLE ~/install-sirf-exercises-dep.py requirements.txt
```

KrisThielemans (changes requested):
the SIRF-exercises dependencies trick seems quite complicated, but if it works for now, let's use it.

If it works, I guess it's time to merge this!

paskino:
closes #735 as it installs CMake 3.25

paskino:
I just noticed that the GPU docker version was built using Ubuntu 22.04 while the non-GPU used Ubuntu 20.04.

I spotted it because the `hdf5-helpers` package came from different repositories: [Ubuntu 22.04 jammy ](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3905695034/jobs/6675520778#step:4:1095)for gpu and [Ubuntu 20.04 focal](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3905695034/jobs/6675520671#step:4:1495) for non-gpu.

Of course, and unfortunately, the Ubuntu 22.04 failed!

paskino:
There are network problems? All jobs fail contacting `archive.ubuntu.com` :(

paskino:
Looks like that in Ubuntu22.04 the `hdf5-helpers` trick does [not work](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/3910975235/jobs/6684605562#step:7:11875).

docker/user_service-ubuntu.sh
```diff
@@ -2,27 +2,46 @@
...
-if [ -f requirements-service.txt ]; then
-  conda install -c conda-forge -y --file requirements-service.txt || \
-  pip install -U -r requirements-service.txt
+if [ "$PYTHON" = "miniconda" ]; then
+  if [ -f requirements-service.yml ]; then
+    # installs the required packages in the environment with requirements-service.yml. 
+    # Notice that these requirements TEMPORARILY contains also the packages for SIRF-Exercises
+    mamba env update --file requirements-service.yml 
+  fi
+  conda clean -y --all
```
KrisThielemans, line 20:
```suggestion
  mamba clean -y --all
```

docker/user_service-ubuntu.sh
KrisThielemans: actually, should this be run in the same virtualenv as in `user_python_ubuntu.sh`?

docker/install-sirf-exercises-dep.py
```diff
@@ -0,0 +1,47 @@
...
+    
+
+    install_by_pip = functools.reduce(lambda x, y: x + ' ' + y, install_by_pip, '')
+    
+
+    try:
+        print ("Installing from conda-forge:", install_by_conda)
+        subprocess.run(conda_install(install_by_conda), shell=True, check=True)
+        print ("Installing from PyPI:", install_by_pip)
+        subprocess.run(pip_install(install_by_pip), shell=True, check=True)
```
KrisThielemans, line 44:
```suggestion
        subprocess.run([sys.executable,  "-m", pip_install(install_by_pip)], shell=True, check=True)
```
and `import sys` above

docker/user_service-ubuntu.sh
```diff
@@ -2,27 +2,46 @@
...
+  fi
 fi
 
 #install SIRF-Exercises requirements
 cd $INSTALL_DIR/SIRF-Exercises
 if [ -f requirements.txt ]; then
-  conda install -c conda-forge -y --file requirements.txt || \
-  pip install -U -r requirements.txt
+  cat requirements.txt
+  python ~/install-sirf-exercises-dep.py requirements.txt
```
KrisThielemans, line 33: This only makes sense when using conda/mamba. Otherwise we just have to use `pip`.
```suggestion
  if [ "$PYTHON" = "miniconda" ]; then
    # WARNING: temporary work-around for requirements not available in conda
    # This will only work when building the docker image as we rely on an absolute path here.
    python ~/install-sirf-exercises-dep.py requirements.txt
  else
    $PYTHON_EXECUTABLE -m pip install -U -r requirements.txt
  fi
```

KrisThielemans (changes requested):
I suggest some changes related to `user_service-ubuntu.sh` when not running `conda/mamba` . Someone might call this outside of the docker environment without conda... (like the other scripts)

paskino:
I ca't believe it's (almost) all green!

KrisThielemans:
Great!

Only failure is due to #770. Can we set that docker GPU job to "allow failure" for the moment?

Clean-up time! Don't forget to use `paralleproj` 1.2.13

KrisThielemans:
Please create an issue for the HDF5 problem. I don't understand why it failed (I suspect it's something to do with cached docker files TBH), but we've run out of time to fix that now.

docker/build_system-ubuntu.sh
KrisThielemans: as it didn't work apparently, let's ignore this comment for now

paskino:
> Please create an issue for the HDF5 problem. I don't understand why it failed (I suspect it's something to do with cached docker files TBH), but we've run out of time to fix that now.

I'm pretty sure it happens on my machine too.

KrisThielemans:
> > Please create an issue for the HDF5 problem. I don't understand why it failed (I suspect it's something to do with cached docker files TBH), but we've run out of time to fix that now.
> 
> I'm pretty sure it happens on my machine too.

but not on the VM.

KrisThielemans:
STIR 5.1.0 is out! Please use that.

docker/user_service-ubuntu.sh
paskino: This can't be done, as `PYTHON_EXECUTABLE` would be either set to `"miniconda"`, if the user wants to use conda, or to the path to the executable otherwise.

In case of using conda, it is safer to use `python` as it would resolve to the current default python executable, which is the miniconda one.

paskino:
I think this PR is ready to be merged. I changed the base branch built in `docker-compose.yml` to `SyneRBI/master`

KrisThielemans (commented):
Looks great. Fantastic!

Can you just update `CHANGES.md` please? Then merge!

paskino:
I did update that, but didn't add to the commit. 
I'll merge now.

Issue #777, opened 2022-11-23
VM: configure_gnome.sh problem on Ubuntu 22.04


KrisThielemans:
I tried to switch to 22.04. All seems well at present (but I haven't got very far yer). However, `configure_gnome.sh` fails to update the GNome settings. This seems due to
https://github.com/SyneRBI/SIRF-SuperBuild/blob/a0ad95c056cd7f666d8e09f296d1234227d50bef/VirtualBox/scripts/configure_gnome.sh#L26-L29
in particular, `pgrep -u gnome-session` returns 3 PIDs. Using `ps` I see they correspond to
```
sirfuser    1375  0.0  0.2 227460 15496 tty2     Sl+  08:26   0:00 /usr/libexec/gnome-session-binary --session=gnome
sirfuser    1423  0.0  0.0  96336  5132 ?        Ssl  08:26   0:00 /usr/libexec/gnome-session-ctl --monitor
sirfuser    1435  0.0  0.2 523372 17552 ?        Ssl  08:26   0:00 /usr/libexec/gnome-session-binary --systemd-service --session=gnome
```
However, if I just comment out those 3 lines, everything works fine (also on our current 18.04 VM, wich doesn't have the equivalent of the 1432 and 1435 processes above).

@paskino, do you remember why we have them? Maybe it was an attempt to get `configure_gnome.sh` in the creation of the VM, but we've abandonded that as it doesn't work anyway.

An old issue somewhat related to this https://github.com/SyneRBI/SyneRBI_VM/issues/164

KrisThielemans:
in an interactive sesoin `DISPLAY` will be set already. Not sure about `*DBUS*` but I recall using `gsetting` in a terminal when the VM is up without any trouble.

paskino:
From my very scientific investigation I currently saw that the 3 PIDs do yield the same result.

KrisThielemans:
but the script is broken, and the simplest way to fix it might be to remove them. anyway. I'll let you get on with it :-)

Issue #768, opened 2022-09-01
update VM to Ubuntu 20.04, or even 22.04


KrisThielemans:

KrisThielemans:
@paskino I think we have to do this now. Is should work, except for #777

PR #793, opened 2023-01-19
Updates to VM for Ubuntu 22.04 base image

paskino:


paskino:
closes #777 
closes #768

PR #797, opened 2023-01-20
install python-is-python3

### LINKED ISSUE #624: find_package(Cython) fails to find cython3
On my Ubuntu 20.04 (under WSL2, but I can't see how that matters) I have `cython3` (I could install `cython` which will give me a Python2 version), but surely don't want to).

Building with `BUILD_CIL_LITE=ON` gives
```
-- SuperBuild -     astra-python-wrapper => Requires astra-toolbox,
-- SuperBuild -       astra-toolbox => Requires Boost[INCLUDED],
-- SuperBuild -       astra-toolbox[OK]
CMake Error at SuperBuild/External_astra-toolbox.cmake:32 (message):
  astra-toolkit binding to Python depends on Cython
  CMake/ExternalProjectDependency.cmake:593 (include)
  SuperBuild/External_astra-python-wrapper.cmake:29 (ExternalProject_Include_Dependencies)
  CMake/ExternalProjectDependency.cmake:593 (include)
  SuperBuild/External_CIL-ASTRA.cmake:28 (ExternalProject_Include_Dependencies)
  CMake/ExternalProjectDependency.cmake:593 (include)
  CMake/ExternalProjectDependency.cmake:639 (ExternalProject_Include_Dependencies)
  SuperBuild.cmake:331 (ExternalProject_Include_Dependencies)
  CMakeLists.txt:43 (include)
```
I can set `CYTHON_EXECUTABLE` to `/usr/bin/cython3`
https://github.com/SyneRBI/SIRF-SuperBuild/blob/053bfabcdfd0967873aaaf68252e7e0ef497cc26/CMake/FindCython.cmake#L11

However, this is probably unexpected for most people. Possibly even worse, it seems that if you have `cython` and its Python2, it'll find that one.

I see that this file comes from https://github.com/symengine/symengine.py/blob/master/cmake/FindCython.cmake which has moved on a bit. However, that one will first find `cython`, which again seems dangerous.

paskino:
closes #624

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,6 @@
 # ChangeLog
 ## v3.4.0
+- Install python-is-python3 deb package in docker/VM.
 - Removed CIL-ASTRA as it has been merged into CIL code base.
 - Added python-opencv to docker requirements.txt
 - docker images updates
--- a/docker/build_python-ubuntu.sh
+++ b/docker/build_python-ubuntu.sh
@@ -3,5 +3,6 @@ set -ev
 apt-get update -qq
 apt-get install -yq --no-install-recommends \
   python3-dev           \
-  python3-tk
+  python3-tk            \
+  python-is-python3
 apt-get clean


KrisThielemans (commented):
I'm somewhat reluctant to do this. It's alright for docker or the VM, but we probably tell people they can execute these scripts if they want on their own system (I often do). This addition seems a bit surprising/dangerous in that situation.

If it solves a problem, then fine, let's merge this, but I'd like to see it documented, at least in the CHANGES

paskino:
I don't think there is a bug per se. However, on Ubuntu 22.04 (our target release) you find, somewhat confusingly, `python3` and `cython`, but not `python` or `cython3`.

I had no problem both on docker and the VM. You reported #624 , for Ubuntu 20.04 which isn't supported.

Up to you if it makes sense to merge. To me it seems a bit more clear.

KrisThielemans:
ok, if we have no problem now, then I wouldn't merge this. I wouldn't want our scripts messing up someone's Ubuntu...

paskino:
They do already, IMHO, if they install all that stuff!

KrisThielemans:
😀 at least, installing extra packages with `apt` shouldn't break anything.

anyway. if it isn't broken now, I wouldn't fix it!

PR #798, opened 2023-01-20
switch to jupyter lab

### LINKED ISSUE #796: switch docker (and VM?) to jupyterlab
The JupyterHub uses jupyterlab. https://github.com/SyneRBI/SIRF-Exercises/pull/189 modifies the files accordingly.
A few places where we still use the old one:

https://github.com/SyneRBI/SIRF-SuperBuild/blob/3c01b373ebb2f7968f93a0ca101c8503f2140f50/docker/service.sh#L61-L62
https://github.com/SyneRBI/SIRF-SuperBuild/blob/91f4f72d19ec68e1efa127e5fc916a531ec2b416/docker/service.multi.sh#L27
https://github.com/SyneRBI/SIRF-SuperBuild/blob/413ff52f309bdcd669e0c27f9c748dde2c71131d/VirtualBox/Using_VM_as_server.md?plain=1#L26
https://github.com/SyneRBI/SIRF-SuperBuild/blob/8dbc041c4ef6dec2255166b665b992f7b699555d/VirtualBox/documentation/Getting-started-with-SIRF.md#L55

I haven't checked if `jupyter lab` works on the new VM, but it does on one that I created ~2 months ago, so we should be fine. @evgueni-ovtchinnikov  can you check?

The docker instance would have to be tested.

paskino:
closes #796

### DIFF
--- a/docker/service.sh
+++ b/docker/service.sh
@@ -51,7 +51,7 @@ fi
 
 echo "start jupyter"
 if [ ! -f ~/.jupyter/jupyter_notebook_config.py ]; then
-  jupyter notebook --generate-config
+  jupyter lab --generate-config
   echo "c.NotebookApp.password = u'sha1:cbf03843d2bb:8729d2fbec60cacf6485758752789cd9989e756c'" \
   >> ~/.jupyter/jupyter_notebook_config.py
 fi
--- a/docker/service.sh
+++ b/docker/service.sh
@@ -50,10 +50,10 @@ if [ ! -r SIRF-contrib ]; then
 fi
 
 echo "start jupyter"
-if [ ! -f ~/.jupyter/jupyter_notebook_config.py ]; then
-  jupyter notebook --generate-config
-  echo "c.NotebookApp.password = u'sha1:cbf03843d2bb:8729d2fbec60cacf6485758752789cd9989e756c'" \
-  >> ~/.jupyter/jupyter_notebook_config.py
+if [ ! -f ~/.jupyter/jupyter_server_config.py ]; then
+  mkdir -p ~/.jupyter/
+  echo "c.ServerApp.password = u'sha1:cbf03843d2bb:8729d2fbec60cacf6485758752789cd9989e756c'" \
+  > ~/.jupyter/jupyter_server_config.py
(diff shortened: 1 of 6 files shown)


docker/service.sh
```diff
@@ -51,7 +51,7 @@ fi
 
 echo "start jupyter"
 if [ ! -f ~/.jupyter/jupyter_notebook_config.py ]; then
-  jupyter notebook --generate-config
+  jupyter lab --generate-config
```
KrisThielemans, line 54: I'm actually not 100% sure why we have this line, as it seems to generate a file with everything commented out anyway.

However, when using `jupyter lab --generate-config`, it generates a file ` /home/sirfuser/.jupyter/jupyter_lab_config.py`, so I think we need to repeat the `echo` line below for that file, such that it implies for both notebook and lab interface

KrisThielemans (changes requested):

KrisThielemans (changes requested):
need to modify `Virtualbox/scripts/configure_jupyter.sh` as well. I have been trying but no success yet.

docker/service.sh
paskino: I don't know. I tried just to launch `jupyter lab` on the VM where I did not update this and I could log in with the normal password.

paskino:
> 

As before, I just issued `jupyter lab` instead of `jupyter notebook` and all worked fine without changing anything you are suggesting.

KrisThielemans:
The password setting is ignored. In fact, we need to create a different file
https://jupyter-server.readthedocs.io/en/stable/operators/migrate-from-nbserver.html

commits coming soon

KrisThielemans:
This should be ready now. I'll probably merge after check complete and trigger a Travis build, but feel free to find problems or merge yourself!

KrisThielemans:
hmmm, Doing a travis build uses ~4800 credits and we have 8500 left. Maybe I should save it for the release itself.

Can you do a local build of the "service" image and see if jupyter works?

KrisThielemans:
I've also get 25000 extra travis credits in the mean time. fast!

Issue #834, opened 2023-08-13
multiple requirements*.*/environment.*


KrisThielemans:
At present, we have multiple occurences of installation of Python dependencies:

Docker (uses mamba):
- [docker/requirements.* in user_python-ubuntu.sh](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/docker/user_python-ubuntu.sh#L55-L67)
- [docker/requirements-service.* in user_service-ubuntu.sh](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/docker/user_service-ubuntu.sh#L14-L27)
- [docker/requirements.* in user_service-ubuntu.sh](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/docker/user_service-ubuntu.sh#L31-L44)
- [SIRF-Exercises/environment.yml or SIRF-Exercises/requirements.txt in user_service-ubuntu.sh](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/docker/user_service-ubuntu.sh#L31-L44)

VM (uses system Python):
- [docker/requirements.txt and docker/requirements-service.txt in INSTALL_python_packages.sh](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/VirtualBox/scripts/INSTALL_python_packages.sh#L77-L78)
- [SIRF-Exercises/requirements.txt in UPDATE.sh](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/VirtualBox/scripts/UPDATE.sh#L130-L132)

Note that our GitHub Actions workflow uses [user_python-ubuntu.sh with system Python](https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/.github/workflows/c-cpp.yml#L96) (to somewhat mimic the VM) and hence uses `docker/requirements.txt`, and nothing from SIRF-Exercises.

[...]

PR #821, opened 2023-07-11
use environment.yml from SIRF-Exercise

paskino:
Uses environment.yml from SIRF-Exercises if using conda, and it is found. Defaults to requirements.txt if not found or not using conda.

closes #814

PR #835, opened 2023-08-15
remove obsolete version restriction for SciPy

KrisThielemans:
temporarily addresses  #834

### DIFF
--- a/docker/requirements.yml
+++ b/docker/requirements.yml
@@ -11,7 +11,7 @@ dependencies:
   - pytest
   - pytest-cov
   - numpy        # CIL
-  - scipy=1.7.3  # CIL prevents `GLIBCXX_3.4.30' not found
+  - scipy
   - docopt
   - matplotlib
   - Cython       # CIL


KrisThielemans:
This built ok, but it seems we're no trunning any CIL tests on the docker images, see https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/5867753842/job/15909452350?pr=835#step:6:25667
that isn't great.

KrisThielemans:
> we're not running any CIL tests on the docker images

This is due to #836.

@paskino @casperdcl @gfardell is there a way to test a conda install of CIL?

casperdcl:
If you mean a one-line `conda install`, then yes... https://github.com/TomographicImaging/CIL#installation-of-cil

casperdcl:
The CIL conda recipe does [include a `test` key](https://github.com/TomographicImaging/CIL/blob/b0503d1b24cc71d937bbb780602d8778b36b0e77/recipe/meta.yaml#L37), so I guess yes... but presumably SIRF only needs to test the points of integration with CIL, not actually run the full CIL test suite.

KrisThielemans:
Well, arguably if you put full `CIL` on our docker image, maybe it should be tested anyway, although I presume that if a conda build it is found&installed, it can be presumed to work?

But certainly, cross-CIL-SIRF tests are the most important.  I'm not sure where they are now. Might need @paskino.

gfardell:
> Well, arguably if you put full CIL on our docker image, maybe it should be tested anyway, although I presume that if a conda build it is found&installed, it can be presumed to work?

We build and test all the binaries we push, this broadly tests numpy and python versions combinations. But we don't have version caps on our dependencies, so when they are updated, conda starts resolving with the newer version and issues can appear due to backward incompatibilities. I think we've had this with SciPy and matplotlib in the past.

To run the tests you need the extra dependencies in the conda `test` requirements that @casperdcl pointed out, but the tests aren't packaged in to the conda build (as far as a recall!) so you'd have to get them from our repo.

KrisThielemans:
@gfardell or anyone else, could you write a small script for CIL testing of an existing installation (i.e. clone, install dependencies, run tests, clean-up). Adding a call to that in the `Dockerfile` would be reassuring.

paskino:
Integration tests with SIRF and CIL are in the CIL test suite https://github.com/TomographicImaging/CIL/blob/master/Wrappers/Python/test/test_SIRF.py

They are not run in CIL CI as we don't have SIRF. They should be run when `BUILD_CIL=ON` but they seem to be commented out!

https://github.com/SyneRBI/SIRF-SuperBuild/blob/50f8a96e45ef6235e46bd237b534a95d122058b1/SuperBuild/External_CIL.cmake#L98-L100

paskino:
@casperdcl can we easily extract the `test` requirements of CIL from the [`meta.yaml`](https://github.com/TomographicImaging/CIL/blob/b0503d1b24cc71d937bbb780602d8778b36b0e77/recipe/meta.yaml#L37-L46)?

paskino:
> @gfardell or anyone else, could you write a small script for CIL testing of an existing installation (i.e. clone, install dependencies, run tests, clean-up). Adding a call to that in the `Dockerfile` would be reassuring.

Isn't it preferrable to `BUILD_CIL=ON` also in Docker? This would simplify this and close #836 after #837

paskino:
I just had a look at the Docker files. I don't think that CIL is installed at all on our docker images, see [`requirements.yaml`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker/requirements.yml) and [`requirements-service.yml`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker/requirements-service.yml) 

It is installed on the [jupyterhub](https://github.com/paskino/SIRF-SuperBuild/blob/jupyterhub_env/jupyterhub/requirements.yml) image via conda.

KrisThielemans:
> I just had a look at the Docker files. I don't think that CIL is installed at all on our docker images, see [`requirements.yaml`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker/requirements.yml) and [`requirements-service.yml`](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/docker/requirements-service.yml)

It is now installed via `SIRF-Exercises/environment.yml`, done in #821 at my request...

> Isn't it preferrable to `BUILD_CIL=ON` also in Docker? This would simplify this and close #836 after #837

I think we have a problem with dependencies as we mix conda and hand-build stuff. If we do `BUILD_CIL=ON`, I believe that we will still install it via `conda` as well, due to the above.

It's of course a mess, which #821 probably made larger. The best way around it is to do everything via `conda` (but we cannot), or nothing... I suggest we merge this and either do 
- a manual check on the docker images
- do minimal work to run some CIL tests automatically on docker
Then we release, and clean it up.

casperdcl:
> can we easily extract the `test` requirements of CIL from the [`meta.yaml`](https://github.com/TomographicImaging/CIL/blob/b0503d1b24cc71d937bbb780602d8778b36b0e77/recipe/meta.yaml#L37-L46)?

yez :)

```sh
CIL$ pip install yq
CIL$ sed s/{{.*}}/./g recipe/meta.yaml | yq -r .test.requires[]
```

paskino (approved):

PR #838, opened 2023-09-12
CI: test docker CIL

casperdcl:
- follow-up to #837
- fixes https://github.com/SyneRBI/SIRF-SuperBuild/pull/835#issuecomment-1709896466
- part of #836
- [x] depends on https://github.com/TomographicImaging/CIL/issues/1509
- is a dependency of https://github.com/SyneRBI/SIRF/issues/1198
- fixes #827

diff: https://github.com/SyneRBI/SIRF-SuperBuild/pull/838/files?w=1

Issue #722, opened 2022-06-09
env files need to set SIRF_PATH to where data is installed


KrisThielemans:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/3950a2a5095cb9d9d66134d83becf186166d3062/env_sirf.sh.in#L6
but this means we need to have access to the source files.

Should instead be something like `${CMAKE_INSTALL_PREFIX}/share/SIRF-3.2/`. see https://github.com/SyneRBI/SIRF-SuperBuild/pull/718/files#r882778578

KrisThielemans:
SIRF finds location for install [here](https://github.com/SyneRBI/SIRF/blob/6fc60a4b81e67f38f8a4343563cc509b0b8ec19e/CMakeLists.txt#L205)

KrisThielemans:
Ideally we should get rid of the use of `SIRF_PATH` at all. Related is https://github.com/SyneRBI/SIRF/pull/904

KrisThielemans:
Pending https://github.com/SyneRBI/SIRF/issues/1201, so postponing

KrisThielemans:
SIRF tests still use the `SIRF_PATH` env variable sadly, see https://github.com/SyneRBI/SIRF/issues/1235, so we cannot remove `SIRF_PATH` yet. However, there are a few things that probably should be resolved now.

Before going into that, note that it is currently set as
https://github.com/SyneRBI/SIRF-SuperBuild/blob/8ca50676b4b3140beb960dc76f2f6ef5d2468ba4/env_sirf.sh.in#L6

Also, I believe that in the new docker env, we don't have the source anymore, so `SIRF_PATH` will point to a non-existent path I suppose. 

## optionally fix now

I found a few places where `SIRF_PATH` is used to find the location of the `gadgetron` executable, but they should all use `SIRF_INSTALL_PATH` instead
https://github.com/SyneRBI/SIRF-SuperBuild/blob/8ca50676b4b3140beb960dc76f2f6ef5d2468ba4/docker/start-gadgetron-notebook.sh#L31-L34
https://github.com/SyneRBI/SIRF-SuperBuild/blob/8ca50676b4b3140beb960dc76f2f6ef5d2468ba4/.github/workflows/test_cil.sh#L12
https://github.com/SyneRBI/SIRF-SuperBuild/blob/8ca50676b4b3140beb960dc76f2f6ef5d2468ba4/docker/devel/test.sh#L12-L15

[...]

PR #846, opened 2023-11-22
re-enable some CIL tests depending on SIRF allocate()

KrisThielemans:
Now that https://github.com/SyneRBI/SIRF/issues/1212 is fixed, we can reenable relevant tests

Reverts https://github.com/SyneRBI/SIRF-SuperBuild/pull/838/commits/160d5e5ac8396efb29322fe354083a7e5b5d8ad8

### DIFF
--- a/.github/workflows/test_cil.sh
+++ b/.github/workflows/test_cil.sh
@@ -10,7 +10,7 @@ for fname in test_SIRF.py utils.py testclass.py; do
 done
 
 $SIRF_PATH/../../INSTALL/bin/gadgetron >& ~/gadgetron.log&
-/opt/conda/bin/python -m unittest discover -v ./cil_sirf_test -k GradientPET -k BlockDataContainer
+/opt/conda/bin/python -m unittest discover -v ./cil_sirf_test
 ret=$?
 for i in $(jobs -p); do kill -n 15 $i || : ; done 2>/dev/null  # kill gadgetron
 exit $ret
--- a/.github/workflows/test_cil.sh
+++ b/.github/workflows/test_cil.sh
@@ -5,6 +5,7 @@ set -euxo pipefail
 mkdir cil_sirf_test
 touch cil_sirf_test/__init__.py
 DEFAULT_CIL_TAG=$(python -c 'import cil; print(f"v{cil.version.version}" if cil.version.commit_hash == "-1" else cil.version.commit_hash[1:])' | tail -n1)
+DEFAULT_CIL_TAG=master  # TODO: revert after CIL 24.0.0 release https://github.com/TomographicImaging/CIL/issues/1732
 for fname in test_SIRF.py utils.py testclass.py; do
   curl -fsSL https://raw.githubusercontent.com/TomographicImaging/CIL/$DEFAULT_CIL_TAG/Wrappers/Python/test/$fname > cil_sirf_test/$fname
 done
--- a/.github/workflows/test_cil.sh
+++ b/.github/workflows/test_cil.sh
@@ -1,16 +1,16 @@
 #!/usr/bin/env bash
-[ -f .bashrc ] && . .bashrc
+[ -f .bashrc ] && . .bashrc # sets paths to gadgetron & python
 set -euxo pipefail
 
 mkdir cil_sirf_test
 touch cil_sirf_test/__init__.py
-DEFAULT_CIL_TAG=$(/opt/conda/bin/python -c 'import cil; print(f"v{cil.version.version}" if cil.version.commit_hash == "-1" else cil.version.commit_hash[1:])' | tail -n1)
+DEFAULT_CIL_TAG=$(python -c 'import cil; print(f"v{cil.version.version}" if cil.version.commit_hash == "-1" else cil.version.commit_hash[1:])' | tail -n1)
 for fname in test_SIRF.py utils.py testclass.py; do
   curl -fsSL https://raw.githubusercontent.com/TomographicImaging/CIL/$DEFAULT_CIL_TAG/Wrappers/Python/test/$fname > cil_sirf_test/$fname
 done
 
-$SIRF_PATH/../../INSTALL/bin/gadgetron >& ~/gadgetron.log&
-/opt/conda/bin/python -m unittest discover -v ./cil_sirf_test
+gadgetron >& ~/gadgetron.log&
+python -m unittest discover -v ./cil_sirf_test
 ret=$?
 for i in $(jobs -p); do kill -n 15 $i || : ; done 2>/dev/null  # kill gadgetron
 exit $ret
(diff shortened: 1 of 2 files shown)


casperdcl:
Based on the failure doesn't look like it's fixed yet /CC @evgueni-ovtchinnikov :sweat_smile:

KrisThielemans:
Not so clear if there's still a SIRF problem. It'll only be the `DEVEL_BUILD=ON` jobs that would work at the moment. That creates suitable complications...

Looks like only one of your CI jobs tests with `DEVEL_BUILD=ON`: https://github.com/SyneRBI/SIRF-SuperBuild/blob/42ce8cbe1e5c18a96e4436fc2d79e87e7accdc73/.github/workflows/c-cpp.yml#L58-L63

Let's hope that https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/6956888225/job/18928662469?pr=846 works therefore, and the docker devel build (but that'll likely get aobrted first if one of the others fails).

In fact, we really should do the `SIRF_CIL` tests are part if the SIRF CI, but they aren't. I suppose we're not building CIL there. Oh well.

PS: By the way, it's a pain to figure out which job is which. we should do something smarter with naming.

evgueni-ovtchinnikov:
I did not expect the use of  `allocate(None)` (looks bizarre to me), quick fix for backward compatibility just pushed.

KrisThielemans:
Unrelated failure in DEVEL_BUILD=ON](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/6956888225/job/18928662469?pr=846#step:11:1095) and the docker devel-service build: https://github.com/TomographicImaging/CIL/issues/1587

KrisThielemans:
By the way, no reason we test the docker CIL tests only in the service images I guess

KrisThielemans:
Possibly needs `version_config.cmake` update for SIRF_TAG to current version, but not sure if that's what is still the problem.

KrisThielemans:
just rebased this. After all the recent changes, maybe it'll work...

KrisThielemans:
Fails with
```
connection to gadgetron server failed, trying again
```
We do try and run it
https://github.com/SyneRBI/SIRF-SuperBuild/blob/8ca50676b4b3140beb960dc76f2f6ef5d2468ba4/.github/workflows/test_cil.sh#L12
it is executed https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/8092243834/job/22112637500#step:10:41, but we don't see the log

Possibly due to #722

KrisThielemans:
@casperdcl to edit `.github/workflows/test_cil.sh`: remove explicit paths and possibly we need to source `env_sirf.sh` first.

casperdcl:
https://github.com/TomographicImaging/CIL/blob/master/Wrappers/Python/test/test_SIRF.py one test failing:

```python
test_TVdenoisingMR (test_SIRF.TestGradientMR_2D) ...
WARNING:root:C backend is only for arrays of datatype float32 - defaulting to `numpy` backend
...
WARNING:root:C backend is only for arrays of datatype float32 - defaulting to `numpy` backend
FAIL
...
======================================================================
FAIL: test_TVdenoisingMR (test_SIRF.TestGradientMR_2D)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/jovyan/cil_sirf_test/test_SIRF.py", line 259, in test_TVdenoisingMR
    np.testing.assert_array_almost_equal(res1.as_array(), res2.as_array(), decimal=3)
Iterations stopped at 100 with the residual 0.005245 
Iterations stopped at 100 with the residual 0.000000 
Iterations stopped at 100 with the residual 0.000500 
  File "/opt/conda/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/opt/conda/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 1099, in assert_array_almost_equal
    assert_array_compare(compare, x, y, err_msg=err_msg, verbose=verbose,
  File "/opt/conda/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/opt/conda/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 862, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 3 decimals

Mismatched elements: 74722 / 131072 (57%)
Max absolute difference: 0.14627695
Max relative difference: 0.3511309
 x: array([[[0.198+0.j, 0.195+0.j, 0.191+0.j, ..., 0.203+0.j, 0.202+0.j,
         0.199+0.j],
        [0.198+0.j, 0.196+0.j, 0.193+0.j, ..., 0.204+0.j, 0.202+0.j,...
 y: array([[[0.196+0.j, 0.195+0.j, 0.193+0.j, ..., 0.204+0.j, 0.202+0.j,
         0.201+0.j],
        [0.196+0.j, 0.195+0.j, 0.194+0.j, ..., 0.203+0.j, 0.202+0.j,...
```

KrisThielemans:
Well, the good thing is that GradientPET BlockDataContainer tests work!

The failing test is https://github.com/TomographicImaging/CIL/blob/7e0291ea455a4aa037900bf248817a539a7b9d67/Wrappers/Python/test/test_SIRF.py#L249-L258
```python
    def test_TVdenoisingMR(self):
        
        # compare inplace proximal method of TV
        alpha = 0.5
        TV = alpha * TotalVariation(max_iteration=10, warm_start=False)
        res1 = TV.proximal(self.image1, tau=1.0)


        res2 = self.image1*0.
        TV.proximal(self.image1, tau=1.0, out=res2)   
        np.testing.assert_array_almost_equal(res1.as_array(), res2.as_array(), decimal=3)
```
Checking the error message, it appears that `res1` and `res2` are almost equal, just not to the required precision. However, that's pretty weird of course. Why would assigning the result be different from storing it in `out`, unless the iterative process didn't converge yet?

@MargaretDuff @paskino @epapoutsellis is this different due to the `warm_start=False` in the first call, but not in the second call? If so, it seems a bug in the test.

MargaretDuff:
>Why would assigning the result be different from storing it in `out`, unless the iterative process didn't converge yet?

With 10 inner iterations, it may not have converged... But then I don't know where the randomness will have come from as the dual in the proximal calculation is allocated with zeros. 


> @MargaretDuff @paskino @epapoutsellis is this different due to the `warm_start=False` in the first call, but not in the second call? If so, it seems a bug in the test.

The `warm_start=False` call is used when TV is defined not when prox is called so this should be fine. 


The TV.proximal is not safe to use in-place at the moment so TV.proximal(x, tau, out=x) is incorrect, but this should be ok here because you use TV.proximal(x, tau, out=x*0)?

casperdcl:
The tests were fixed in CIL but not yet released. Pushed a temp fix for now, to be reverted after https://github.com/TomographicImaging/CIL/issues/1732.

I suggest we merge this & open a follow-up issue to track https://github.com/TomographicImaging/CIL/issues/1732.

PR #854, opened 2024-02-01
Add explicit CMake version check in case of Gadgetron testing ON

### LINKED ISSUE #823: building Gadgetron with testing requires CMake 3.23 (missing target GTest::gmock)
With CMake 3.17 we get
```
GTEST FOUND: TRUE
...
-- Configuring done
CMake Error at test/CMakeLists.txt:62 (add_executable):
  Target "test_all" links to target "GTest::gmock" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```
https://cmake.org/cmake/help/latest/module/FindGTest.html states that `GTest::gmock` is introduced in CMake 3.23.

We should therefore check CMake version and set `BUILD_TESTING_Gadgetron=OFF` and add a warning if CMake is too old to enhance our user experience.

paskino:
Adds an explicit check of CMake version and `message(FATAL_ERROR)` if CMake version is below 3.23.

Closes #823

### DIFF
--- a/SuperBuild/External_Gadgetron.cmake
+++ b/SuperBuild/External_Gadgetron.cmake
@@ -137,6 +137,9 @@ endif()
 
   # Gadgetron only adds tests if (GTEST_FOUND AND ARMADILLO_FOUND)
   if (BUILD_TESTING_${proj})
+    if(CMAKE_VERSION VERSION_LESS "3.23") 
+       message(FATAL_ERROR "You need at least CMake 3.23 to build Gadgetron tests.")
+    endif()
     add_test(NAME ${proj}_TESTS
          COMMAND ${CMAKE_CTEST_COMMAND} -C $<CONFIGURATION> --output-on-failure
          WORKING_DIRECTORY ${${proj}_BINARY_DIR}/test)


KrisThielemans:
The job is going through, but that'll be because CMake is more recent presumably, so you can just as well keep an job with the testing on.

To test if the lines are ok, I guess you'll need an older CMake...

KrisThielemans:
confirmed with CMake 3.10
```
Make Error at SuperBuild/External_Gadgetron.cmake:141 (message):
  You need at least CMake 3.23 to build Gadgetron tests.
```

KrisThielemans:
sadly unrelated error https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/7740032215/job/21104114360?pr=854#step:8:9340
```
CMake Error at /home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/ROOT/build/TBB-prefix/src/TBB-stamp/TBB-build-Release.cmake:49 (message):
  Command failed: 2

   'make' 'compiler=gcc' 'cpp0x=1' 'CXXFLAGS= -mno-rtm' 'CPLUS=/usr/lib/ccache/g++-11' 'CONLY=/usr/lib/ccache/gcc-11' 'LDFLAGS='
  See also
    /home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/ROOT/build/TBB-prefix/src/TBB-stamp/TBB-build-*.log
```
No idea about that.

As it's unrelated, I vote for merging this PR anyway, after cleaning up the jobs.

PR #856, opened 2024-02-01
default STIR_BUILD_SWIG_PYTHON=ON

KrisThielemans:
It's nice to be able to use STIR Python for things that are not in `stir.STIR`, such as examining/manipulating projection data headers.

This PR increases build time a bit, but is otherwise harmless. Would need addition to `CHANGES.md`.

Any objections?


KrisThielemans:
one job failed due to #861

PR #872, opened 2024-02-29
Use OpenMP_<lang>_INCLUDE_DIR as opposed to OPENMP_INCLUDE_DIRS

### LINKED ISSUE #870: parallelproj build failed on OSX due to omp.h not found
@francescaleek has a failure funding `omp.h` when building parallelproj, but not any other project (whcih is a bit weird, but could be because STIR does not require OpenMP). The parallelproj CMake settings look like this:
image

We currently do
https://github.com/SyneRBI/SIRF-SuperBuild/blob/e43323735471174437666bf4134721e1413529ad/SuperBuild/External_parallelproj.cmake#L43-L48
similarly for STIR
https://github.com/SyneRBI/SIRF-SuperBuild/blob/e43323735471174437666bf4134721e1413529ad/SuperBuild/External_STIR.cmake#L105-L106
and overall (as [`mark_as_superbuild` should transport those to the subprojects](https://cmake-artichoke.readthedocs.io/en/latest/ExternalProjectDependency.html))
https://github.com/SyneRBI/SIRF-SuperBuild/blob/e43323735471174437666bf4134721e1413529ad/SuperBuild.cmake#L225-L231

I note we set `OPENMP_INCLUDES`, but https://cmake.org/cmake/help/latest/module/FindOpenMP.html says we should use `OpenMP_CXX_INCLUDE_DIR` (and some for C)

Note that none of this is necessary on Linux/g++ but @rijobro put it in for MacOS. Possibly `FindOpenMP.cmake` has moved on.

KrisThielemans:
Necessary on MacOS.

Fixes #870


KrisThielemans:
@francescaleek, could you give this a go?

First find out where your `omp.h` is (not sure how to do that on a Mac), most likely `/usr/local/include`, then do something like this for an existing build
```
cd SIRF-SuperBuild
git remote add kris https://github.com/KrisThielemans/SIRF-SuperBuild.git
git fetch kris
git checkout OpenMP_fixes
cd ../build
cmake -DOpenMP_CXX_INCLUDE_DIR:PATH=/usr/local/include -DOpenMP_C_INCLUDE_DIR:PATH=/usr/local/include .
make -j4
```
you'd have to reenable parallelproj and NiftyReg OpenMP to see of course.

If you have a working build already, you could copy all of that first, such that you don't screw it up.

It's not urgent to test this, but if you can and it works, then we'll include it in the imminent release.

KrisThielemans:
This seems to have worked mostly for @francescaleek, except for NiftyReg. I've created https://github.com/KCL-BMEIS/niftyreg/pull/110. We could update `version_config.cmake` to point to that branch, or presumably set in `External_NIFTYREG.cmake`
```cmake
set(OpenMP_C_FLAGS "${OpenMP_C_FLAGS} -I${OpenMP_C_INCLUDE_DIRS")
set(OpenMP_CXX_FLAGS "${OpenMP_CXX_FLAGS} -I${OpenMP_CXX_INCLUDE_DIRS")
```
maybe pointing to a correct version is best. @paskino @casperdcl any opinion?

KrisThielemans:
@francescaleek, the NiftyReg fix has been merged, so can you try again with
```sh
cmake -DOpenMP_CXX_INCLUDE_DIR:PATH=/usr/local/include -DOpenMP_C_INCLUDE_DIR:PATH=/usr/local/include -DIFTYREG_TAG:STRING=a328efb3a2f8ea4b47cf0f7b581d983a570a1ffd .
make -j4
```
Hopefully this does the trick...

PR #873, opened 2024-02-29
workflow tidy

### LINKED ISSUE #860: GHA: need to use checkout:v4
we're getting warnings about nodeJS 16. I guess upgrading the action will fix this

casperdcl:
`c-cpp.yml` (`native`) and `docker.yml` workflow updates

- shorter job names
- bump actions (fixes #860)
- fix `tmate` debugging
- cache `pip`
- `ccache` fallback keys
- better debug artefact name
- fix docker release auth
- use https://github.com/jlumbroso/free-disk-space action
- fix silent failures

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,5 +1,4 @@
-name: run superbuild and test
-
+name: native
 on:
   push:
     branches: [ master ]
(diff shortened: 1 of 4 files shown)


.github/workflows/c-cpp.yml
```diff
@@ -25,140 +24,124 @@ on:
...
+        default: false
+        required: true
 jobs:
-  build:
-
+  build-test:
     runs-on: ${{ matrix.os }}
+    defaults: {run: {shell: bash}}
+    env:
+      COMPILER: gcc
```
KrisThielemans, line 39: can we still override below? we had clang jobs at some point...

casperdcl: we can; but we weren't using them atm

PR #874, opened 2024-02-29
docker: USE_SYSTEM_HDF=ON

### LINKED ISSUE #808: have to use USE_SYSTEM_HDF=OFF on docker
@paskino saw this when releasing SIRF 3.4. My diagnosis is at
https://github.com/SyneRBI/SIRF-SuperBuild/pull/787#issuecomment-1379254824
and essentially says the system hdf5 on Ubuntu doesn't come with the cmake export files (why? Who knows) and you need to install h5cc to get it to work, which implies installing hdf5-helpers.

However, Edo tried this and said it didn't work
https://github.com/SyneRBI/SIRF-SuperBuild/pull/787#issuecomment-1381986403
so he gave up and set USE_SYSTEM_HDF5=OFF https://github.com/SyneRBI/SIRF-SuperBuild/blob/79cc17a0b11228a3cb40951434e8d9643824735d/docker/Dockerfile#L139.

If we build HDF5 ourselves, then the cmake export files are there, and everything works fine.
This isn't great of course as it adds to our building time  and docker size etc.

I haven't investigated further why installing hdf5-helpers didn't work on docker.

casperdcl:
- fixes #808


casperdcl:
whelp that was easy 😅

PR #876, opened 2024-03-04
VM update script updates

### LINKED ISSUE #869: update_VM.sh will fail due to STIR incompatibility
Previous versions of STIR wrote files in `${CMAKE_INSTALL_PREFIX}/include/stir` and `${CMAKE_INSTALL_PREFIX}/lib/cmake/STIR*cmake`. STIR 6.0 writes them in version specific folders https://github.com/UCL/STIR/pull/1358. However, an existing config of SIRF seems to still find the old ones. This can be resolved by (for instance on the VM)
```sh
cd ~/devel/install
rm -rf include/stir
rm -f lib/STIR*.cmake
```
before `updateVM.sh`.

We could add this to the `UPDATE.sh` script, or point users to this in the doc (although who reads that?)

In fact, we're going to have lots of problems like this, so I suggest to nuke `~/devel/install` as well as `~/devel/buildVM/builds/`.

@casperdcl @paskino what do you think?

KrisThielemans:
Fixes #869

### DIFF
--- a/VirtualBox/documentation/Basic-usage.md
+++ b/VirtualBox/documentation/Basic-usage.md
@@ -86,9 +86,12 @@ We  provide a script to update an existing VM to a new version of course softwar
 
 If you have decided to upgrade your existing VM, type the following in a terminal
 
+     cd ~/devel
+     rm -rf install/lib install/include
+     rm -rf buildVM
      update_VM.sh -s
 
-The `-s` option will update your system dependencies via APT.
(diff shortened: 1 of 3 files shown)


KrisThielemans:
I've chosen the safe route here, and just recommend to delete more files. We could instead do this by default and add an option (e.g. '--preserve') for developers to prevent that, but I don't have time for that, and neither do you. So I suggest to merge this.

casperdcl (approved):
`disown $(pidof virtualbox)`; ping @paskino for review

KrisThielemans:
I'm going to merge this. If we experience problems when building the VM, we'll fix it then.

PR #879, opened 2024-03-09
pass through OpenMP_iomp_LIBRARY

### LINKED ISSUE #502: Why are OpenMP flags and directories passed to external projects?
I am using the intel compiler and I cannot build for instance parallelproj with OpenMP because we are passing somehow the flags and libraries found to external project via [these lines](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/SuperBuild.cmake#L230L236)

```cmake

mark_as_superbuild(ALL_PROJECTS VARS 
      OpenMP_CXX_FLAGS:STRING OpenMP_CXX_LIB_NAMES:STRING OpenMP_C_FLAGS:STRING 
      OpenMP_C_LIB_NAMES:STRING OpenMP_CXX_FLAGS:STRING OpenMP_CXX_LIB_NAMES:STRING 
      OpenMP_libomp_LIBRARY:FILEPATH OpenMP_omp_LIBRARY:FILEPATH
      OpenMP_gomp_LIBRARY:FILEPATH OpenMP_pthread_LIBRARY:FILEPATH
      OPENMP_INCLUDES:PATH OPENMP_LIBRARIES:PATH
      )
```

This seems to prevent the project's own `find_package(OpenMP)` to set the variable it requires, probably because `find_package` looks in the cache before executing. I get the a message that says that the `OpenMP_iomp_LIBRARY` variable isn't set and the build of the external project fails at cmake configure time. 

The solution is to add `OpenMP_iomp_LIBRARY` to the variables in `mark_as_superbuild`. However, `OpenMP_iomp_LIBRARY` is a variable that has meaning only if one tries to build with the intel compiler, while the others seem to be applying to GCC. So, while adding the variable solves the issue, I don't think we should maintain such a list at all.

Actually, removing the above code in ~~`mark_as_advanced`~~ `mark_as_superbuild` makes it work as well.

So the question is why do we pass such variables at all?

KrisThielemans:
also only pass those `OpenMP*LIBRARY` that have been set.

Fixes #502


KrisThielemans:
This works fine, but if you'd prefer to postpone this till after 3.6, that's fine for me.

PR #884, opened 2024-03-10
[VB] add extra options for developers

KrisThielemans:
allow setting DEVEL_BUILD=ON and skipping the build. The latter is mostly useful when creating a VM.


KrisThielemans:
I've tested this locally so will just merge. sorry.

PR #891, opened 2024-03-19
docker docs

### LINKED ISSUE #889: Another Docker README question http or https for localhost?
Seems to be that I have to use http and not https for the localhost link. Am I doing things correctly?

### LINKED ISSUE #890: available docker tags (no devel?)
I have no idea which tags are available, and it seems ghcr.io doesn't have a search/list facility (or not an obvious one anyway). 

The README mentions `jupyter` and `jupyter-gpu`, but there is more. 

I see `core` (but maybe this is internal and doesn't need to be documented)
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fd4e9720965f846889e468eb938c88e5805ff5d3/.github/workflows/docker.yml#L81
although the tag might be different (a bug?)
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fd4e9720965f846889e468eb938c88e5805ff5d3/.github/workflows/docker.yml#L94-L95

There's a 2 push statements
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fd4e9720965f846889e468eb938c88e5805ff5d3/.github/workflows/docker.yml#L107
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fd4e9720965f846889e468eb938c88e5805ff5d3/.github/workflows/docker.yml#L112

I see `jupyter-devel` being created
https://github.com/SyneRBI/SIRF-SuperBuild/blob/fd4e9720965f846889e468eb938c88e5805ff5d3/.github/workflows/docker.yml#L176
But I cannot find a `push` for the `devel` image.

Indeed, trying to create a Codespace using `jupyter-devel` fails https://github.com/SyneRBI/SIRF-Exercises/blob/devel-container/.devcontainer/devcontainer.json
```
2024-03-18 10:37:06.280Z: Creating container...
2024-03-18 10:37:06.330Z: $ devcontainer up --id-label Type=codespaces --workspace-folder /var/lib/docker/codespacemount/workspace/SIRF-Exercises --mount type=bind,source=/.codespaces/agent/mount/cache,target=/vscode --user-data-folder /var/lib/docker/codespacemount/.persistedshare --container-data-folder .vscode-remote/data/Machine --container-system-data-folder /var/vscode-remote --log-level trace --log-format json --update-remote-user-uid-default never --mount-workspace-git-root false --omit-config-remote-env-from-metadata --skip-non-blocking-commands --skip-post-create --config "/var/lib/docker/codespacemount/workspace/SIRF-Exercises/.devcontainer/devcontainer.json" --override-config /root/.codespaces/shared/merged_devcontainer.json --default-user-env-probe loginInteractiveShell --container-session-data-folder /workspaces/.codespaces/.persistedshare/devcontainers-cli/cache --secrets-file /root/.codespaces/shared/user-secrets-envs.json
2024-03-18 10:37:06.499Z: @devcontainers/cli 0.56.1. Node.js v18.19.0. linux 6.2.0-1019-azure x64.
2024-03-18 10:37:08.312Z: {"outcome":"error","message":"Command failed: docker inspect --type image ghcr.io/synerbi/sirf:jupyter-devel","description":"An error occurred setting up the container."}
2024-03-18 10:37:08.314Z: Error: Command failed: docker inspect --type image ghcr.io/synerbi/sirf:jupyter-devel
```

https://github.com/SyneRBI/SIRF-SuperBuild/tree/master/docker#docker-stacks mentions `synerbi/jupyter:sirf` but also `synerbi/sirf:latest`, which isn't documented anywhere?

By the way, 
https://hub.docker.com/r/synerbi/sirf/tags? gives me mostly stuff that is 4 months old, aside from `edge` and `edgegpu`. I have no clue what those are.

Looks like, we have to update our `CHANGES.md` to say that we no longer push to DockerHub and people need to read new instructions.

casperdcl:
- fix `localhost` URL (fixes #889)
- fix `-gpu` suffix
- document tags (fixes #890)
  + drop `:jupyter` tag (just use `:latest`) [5168369](https://github.com/SyneRBI/SIRF-SuperBuild/pull/891/commits/5168369b87bd988f3d7b5ce43849b40eb9200f94) - maybe revert?

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -19,7 +19,7 @@ docker run --rm -it -p 9999:8888 --gpus all ghcr.io/synerbi/sirf:jupyter-gpu
 > [!TIP]
 > docker tag | CIL branch/tag
 > :---|:---
-> `latest`, `latest-gpu` | [latest tag `v*.*.*`](https://github.com/SyneRBI/SIRF-SuperBuild/releases/latest)
+> `latest`, `latest-gpu`, `jupyter`, `jupyter-gpu` | [latest tag `v*.*.*`](https://github.com/SyneRBI/SIRF-SuperBuild/releases/latest)
 > `M`, `M.m`, `M.m.p`, `M-gpu`, `M.m-gpu`, `M.m.p-gpu` | tag `vM.m.p`
 > `edge`, `edge-gpu` | `master`
 > only build & test (no tag) | CI (current commit)
--- a/docker/README.md
+++ b/docker/README.md
@@ -17,15 +17,19 @@ docker run --rm -it -p 9999:8888 --gpus all ghcr.io/synerbi/sirf:latest-gpu
 ```
 
 > [!TIP]
-> docker tag | CIL branch/tag
+> docker tag | git branch/tag
 > :---|:---
 > `latest`, `latest-gpu` | [latest tag `v*.*.*`](https://github.com/SyneRBI/SIRF-SuperBuild/releases/latest)
 > `M`, `M.m`, `M.m.p`, `M-gpu`, `M.m-gpu`, `M.m.p-gpu` | tag `vM.m.p`
 > `edge`, `edge-gpu` | `master`
-> only build & test (no tag) | CI (current commit)
 > `devel`, `devel-gpu` | `master` with `cmake -DDEVEL_BUILD=ON -DBUILD_CIL=ON`
 >
 > See [`ghcr.io/synerbi/sirf`](https://github.com/SyneRBI/SIRF-SuperBuild/pkgs/container/sirf) for a full list of tags.
+>
+> The [`docker.yml` workflow](../.github/workflows/docker.yml) builds & pushes all the docker tags above.
+> Additionally, `core` & `core-gpu` intermediate (cache) docker tags are built & pushed by the workflow, but are not intended for users.
+
+The workflow will also build & test all PRs (without pushing any new image tags).
 
 The Jupyter notebook should be accessible at <http://localhost:9999>.
 
@@ -165,3 +169,6 @@ Note that this cache is different from the "normal" `ccache` of your host. (If y
   + Install [`requirements.yml`](requirements.yml)
   + Clone & setup <https://github.com/SyneRBI/SIRF-Exercises> & <https://github.com/TomographicImaging/CIL-Demos>
   + Set some environment variables (e.g. `PYTHONPATH=/opt/SIRF-SuperBuild/INSTALL/python`, `OMP_NUM_THREADS=$(( cpu_count/2 ))`)
+
+> [!NOTE]
+> `synerbi/jupyter:*` are only intermediate (cache) images not intended for users.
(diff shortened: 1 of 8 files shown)


casperdcl:
@KrisThielemans I prefer the last commit here, but up to you if you'd prefer to revert it

KrisThielemans:
I think this table helps a lot! Thanks. Comments on the table:
- I think the row `only build & test (no tag) | CI (current commit)` is a bit confusing though. Maybe this should be `built & tested during CI (no tag) | master` ? (maybe it should be in bold?)
- The first tag you see at https://github.com/SyneRBI/SIRF-SuperBuild/pkgs/container/sirf is `core` but it's not in the table, which probably is confusing.

I don't have a strong opinion on `latest` vs `jupyter`. I have no idea what conventions people generally use for docker tagging. We had `sirf:latest` and it's a reasonable name, more so than `sirf:jupyter`, so I'm fine with keeping `latest`.

I note that we had `sirf:latest` and `sirf:service`. We no longer have the latter, and I think that's fine, as it was a bit confusing. I'm assuming (in fact, I know) that `sirf:latest` now has the jupyter service running, so we could easily claim that these 2 have been merged into 1. (I do think that this needs to be stated explicitly in `CHANGES.md`).

I'm still confused though by the last bit of the README, which talks about `synerbi/jupyter:sirf` images made by copying stuff from `sirf:latest`. Is that correct? Shouldn't they be in the table then? But they are not pushed?

docker/README.md
```diff
@@ -11,20 +11,31 @@ The image contains SIRF & all dependencies required by JupyterHub.
...
-docker run --rm -it -p 9999:8888 ghcr.io/synerbi/sirf:jupyter
+docker run --rm -it -p 9999:8888 ghcr.io/synerbi/sirf:latest
 # GPU version
-docker run --rm -it -p 9999:8888 --gpus all ghcr.io/synerbi/sirf:jupyter-gpu
+docker run --rm -it -p 9999:8888 --gpus all ghcr.io/synerbi/sirf:latest-gpu
 ```
 
-The Jupyter notebook should be accessible at <https://localhost:9999>.
+> [!TIP]
+> docker tag | CIL branch/tag
```
casperdcl, line 20:
```suggestion
> docker tag | git branch/tag
```

docker/README.md
```diff
@@ -11,20 +11,31 @@ The image contains SIRF & all dependencies required by JupyterHub.
...
+> :---|:---
+> `latest`, `latest-gpu` | [latest tag `v*.*.*`](https://github.com/SyneRBI/SIRF-SuperBuild/releases/latest)
+> `M`, `M.m`, `M.m.p`, `M-gpu`, `M.m-gpu`, `M.m.p-gpu` | tag `vM.m.p`
+> `edge`, `edge-gpu` | `master`
+> only build & test (no tag) | CI (current commit)
```
casperdcl, line 25:
```suggestion
```

casperdcl (commented):
`synerbi/jupyter:*` and `synerbi/sirf:core*` are simply internal caches, so aren't in the table.

KrisThielemans:
> `synerbi/jupyter:*` and `synerbi/sirf:core*` are simply internal caches, so aren't in the table.

ok. Maybe a line with a clarification then for people like me ? :smile:

I'm fine with your other suggestions.

Issue #899, opened 2024-05-14
ROOT: build error on GHA


paskino:
Currently [builds fail](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/9070313125/job/24921825770#step:10:13597) with `BUILD_ROOT=ON` . This seems to be a transient error due to gremlins.

See:
- https://github.com/SyneRBI/SIRF-SuperBuild/pull/897#issuecomment-2109543050
- https://github.com/root-project/root/issues/14520

```
CMake Error at /home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/ROOT/build/builtins/xrootd/XROOTD-prefix/src/XROOTD-stamp/XROOTD-install-Release.cmake:49 (message):
  Command failed: 1

   '/usr/local/bin/cmake' '-Dmake=' '-Dconfig=' '-P' '/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/ROOT/build/builtins/xrootd/XROOTD-prefix/src/XROOTD-stamp/XROOTD-install-Release-impl.cmake'
```

### Suggested fix

https://github.com/SyneRBI/SIRF-SuperBuild/pull/898#issuecomment-2111226591

>I don't think we can have hinds falling regularly. But it's not great to have a lot of jobs falling silently either. So, I go back to my suggestion to install ROOT in another way for most jobs. Possibly, this is already done for the docker builds.

PR #903, opened 2024-05-30
attempt to fix ROOT building on GHA by non-parallel build

KrisThielemans:
let's see if #899 is due to parallel build


KrisThielemans:
that did work, but results in ~4 hour build time

KrisThielemans:
Potentially could re-use this idea when resolving #904

PR #912, opened 2024-06-28
build cil

### LINKED ISSUE #911: docker: edge images have old CIL
[CIL v24.0.0](https://github.com/TomographicImaging/CIL/releases/tag/v24.0.0) from 2 months ago via conda - rather than [CIL master](https://github.com/TomographicImaging/CIL) via CMake/pip - is installed in [ghcr.io/synerbi/sirf:edge](https://github.com/SyneRBI/SIRF-SuperBuild/pkgs/container/sirf) images.

Unfortunately the careful handling of `BUILD_CIL` in #897 seems not to have been enough.

casperdcl:
- fix: trigger `BUILD_CIL` logic in `./docker/compose.sh -d`
- always `BUILD_CIL=ON` for docker `edge,latest,M.m.p` tags (fixes #911)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -88,7 +88,7 @@ jobs:
       name: build
       run: |
         # rebuild sirf image
-        ./docker/compose.sh -bR${{ matrix.type == 'gpu' && 'g' || 'c' }}
+        ./docker/compose.sh -bR${{ matrix.type == 'gpu' && 'g' || 'c' }} -- --build-arg BUILD_CIL=ON
         image=synerbi/sirf:latest${{ steps.pull.outputs.suffix }}
         echo "image=$image" >> "$GITHUB_OUTPUT"
         # tag potentially newer core image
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       name: build
       run: |
         # rebuild sirf image
-        ./docker/compose.sh -bR${{ matrix.type == 'gpu' && 'g' || 'c' }}
+        ./docker/compose.sh -bR${{ matrix.type == 'gpu' && 'g' || 'c' }} -- --build-arg BUILD_CIL=ON
```
casperdcl, line 91: if we only want BUILD_CIL for `edge` but not `latest,M.m.p` then...

```suggestion
        ./docker/compose.sh -bR${{ matrix.type == 'gpu' && 'g' || 'c' }} ${{ github.ref == 'refs/heads/master' && '-- --build-arg BUILD_CIL=ON' || '' }}
```

KrisThielemans (commented):
For me, we should always use BUILD_CIL=ON. That way, we always honour the relevant tag in version_config.cmake..

If this is what did PR does, then I approve!

PR #930, opened 2024-09-03
update SIRF_TAG to hash of 3Sep2024

KrisThielemans:


casperdcl (approved):
pinned https://github.com/SyneRBI/PETRIC-backend/blob/main/Dockerfile so fine to merge

Issue #932, opened 2024-10-02
Error building Tomophantom


NicoleJurjew:
When I try to build with build CIL = ON, I get an error. This seems to be from Tomophantom.

KrisThielemans:
The error said something about `cannot find msvc compiler`...

I see that Tomophantom has move on dramatically now. Should we update it? Or should we even include it? Currently, it is enabled via `BUILD_CIL`. Might need a separate `BUILD` variable then.

paskino:
But we are using a specific version, despite the package changes.
https://github.com/SyneRBI/SIRF-SuperBuild/blob/5f1a7f498bf6f41551c19eb53798b5d33af43eec/version_config.cmake#L127

KrisThielemans:
We do have a frozen version, which is good, but it is old and OS/compilers etc move on, which possibly caused this problem.

I'm not sure how easy it will be for @NicoleJurjew to reproduce this now (and how much time she has for it!). @NicoleJurjew can you just give OS, CMake and gcc version?

In any case, I think it'd be good to be able to switch-off building of TomoPhantom, independent of CIL.

paskino:
Should we discuss whether to include TomoPhantom and CCPi-Regularisation as dependencies of CIL? 
https://github.com/SyneRBI/SIRF-SuperBuild/blob/c5251d17dc2a4a612bc4b4247b07704d4ccb448b/SuperBuild.cmake#L314

Potentially we could add a `BUILD_TomoPhantom` etc to force the build.

KrisThielemans:
sure. I thought you needed it for the CIL-demos, but by all means, make it optional.

Issue #313, opened 2020-01-26
Need to switch to use CMake's FindPython


KrisThielemans:
See https://github.com/CCPPETMR/SIRF/issues/504

For the SuperBuild, there's an additional problem that not all projects will use [`FindPython`](https://cmake.org/cmake/help/v3.12/module/FindPython.html), so somehow we'll have to pass `${Python_EXECUTABLE}` to `PYTHON_EXECUTABLE`. It's going to be messy...

KrisThielemans:
SIRF PR was https://github.com/SyneRBI/SIRF/pull/868

KrisThielemans:
Create `Python_CMAKE_ARGS` and probably define both `Python_EXECUTABLE` to `PYTHON_EXECUTABLE`

KrisThielemans:
This is now pretty urgent, as recent CMake claims this support will be removed.

I've checked for remaining packages using the outdated `FindPythonInterp.cmake`. There's only 2
- https://github.com/TomographicImaging/CIL/issues/1835
- we're using TomoPhantom v2.0.0, which has that problem in its https://github.com/dkazanc/TomoPhantom/blob/v2.0.0/Wrappers/Python/CMakeLists.txt#L14 However, that whole directory no longer exists (v3.0 removed Cython wrappers), so presumably we could upgrade? @gfardell @paskino ?

paskino:
In CIL we switched to `find_python(Python COMPONENTS Interpreter)` and CMake stopped finding the right python. Long story short, we need to pass the path to `Python_Executable`, see

* https://github.com/TomographicImaging/CIL/issues/2040

KrisThielemans:
yes, of course. Once there are multiple Pythons around, it seems entirely down to luck if CMake finds the one you want. I'd assume it first tries "python" in your (shell) path, which would be fine in an activated conda env, but I don't know if that's really true, so generally recommend `-DPython_EXECUTABLE=`which python``. This is true for both the outdated and "new" style `find_package()` way to find python.

It's good to hear that CIL moved on. Maybe we should still cope with older CIL versions in the SB?

What about TomoPhantom? (Personally, I'd suggest to remove this from the default build, if it's still present, that is.)

PR #933, opened 2024-10-02
Updated modules in find_package command for Python3

NicoleJurjew:


KrisThielemans:
https://cmake.org/cmake/help/latest/module/FindPythonInterp.html deprecated since 3.12, while we require 3.16.2.

Current change is minimal. We probably should pass `Python_EXECUTABLE` as opposed to `PYTHON_EXECUTABLE` to dependencies, but this needs a check which ones we use. Probably easiest to pass both. I think we should no longer attempt to pass include and other paths, and let CMake sort it out.

SuperBuild.cmake
```diff
@@ -47,7 +47,7 @@ endif()
 # Find CUDA
 option(DISABLE_CUDA "Disable CUDA" OFF)
 if (NOT DISABLE_CUDA)
-  find_package(CUDA)
+  find_package(CUDAToolkit REQUIRED)
```
KrisThielemans, line 50:
```suggestion
  find_package(CUDA)
```
While this is needed, let's do it in another PR, as both of these could create discussions.

SuperBuild.cmake
```diff
@@ -96,23 +96,23 @@ if (DISABLE_PYTHON)
...
-  find_package(PythonLibs 3)
-  
   if (PYTHONLIBS_FOUND)
     message(STATUS "Found PYTHON_INCLUDE_DIRS=${PYTHON_INCLUDE_DIRS}")
     message(STATUS "Found PYTHON_LIBRARIES=${PYTHON_LIBRARIES}")
   endif()
 
   # Set destinations for Python files
-  set (BUILD_PYTHON ${PYTHONLIBS_FOUND})
+  set (BUILD_PYTHON = ${Python3_FOUND})
```
KrisThielemans, line 115:
```suggestion
  set (BUILD_PYTHON ${Python3_FOUND})
```

KrisThielemans:
Need to update how we look for cython I guess
```
[100%] Generating build/timestamp
Traceback (most recent call last):
  File "/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/TomoPhantom/build/Wrappers/Python/setup.py", line 19, in <module>
    from Cython.Distutils import build_ext
ModuleNotFoundError: No module named 'Cython'
gmake[5]: *** [Wrappers/Python/CMakeFiles/PythonWrapper.dir/build.make:76: Wrappers/Python/build/timestamp] Error 1
gmake[4]: *** [CMakeFiles/Makefile2:157: Wrappers/Python/CMakeFiles/PythonWrapper.dir/all] Error 2
gmake[3]: *** [Makefile:136: all] Error 2
gmake[2]: *** [CMakeFiles/TomoPhantom.dir/build.make:86: builds/TomoPhantom/stamp/TomoPhantom-build] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:755: CMakeFiles/TomoPhantom.dir/all] Error 2
```
This error is different from why I saw in  #932 but it's possible that @NicoleJurjew only saw the Tomophantom error when using this branch I guess.

KrisThielemans:
that's a bit strange. What is the content of env_sirf.sh etc then? Can you just do a
```cmake
message(STATUS "PYTHON_EXECUTABLE=${PYTHON_EXECUTABLE}")
```
before those statement to check what it is?

NicoleJurjew:
This is the content of env_sirf.sh:
grafik

Where will that message be written to then? Thanks!

paskino:
See also #313

casperdcl:
Looks  like a gadgetron build error related to https://groups.google.com/g/linux.debian.bugs.dist/c/JpHAw3GY0Bk

KrisThielemans:
This was all discussed in #937. I've contributed a PR to Gadgetron. However, moving on with Gadgetron created a lot of other problems. So @paskino and I decided to wait with that, and just fix Ubuntu 22.04. This is done in #943. So suggestion therefore is to finish #943, merge it, and merge that back on here.

KrisThielemans (commented):
Did you check if the dependencies indeed accept `Python3_EXECUTABLE`. In any case, I'm fairly sure this needs to be `Python_EXECUTABLE`. Easiest might be to pass both of course, although that's a bit ugly.

Also, in SIRF and STIR, we have a "backwards compatible" line to set `Python_EXECUTABLE` if `PYTHON_EXECUTABLE` is set. Maybe no longer need it?

KrisThielemans:
@casperdcl I've merged #943.

Note that we'll also need to update CHANGES.md here.

SuperBuild.cmake
casperdcl, line 1: technically

`find_package` | `-D` hint
--|--
`Python3` | `Python3_EXECUTABLE`
`Python` | `Python_EXECUTABLE`

so perhaps we could just use `Python` without the `3`-suffix everywhere?

SuperBuild.cmake
```diff
@@ -96,23 +96,21 @@ if (DISABLE_PYTHON)
...
     set(Python_ADDITIONAL_VERSIONS ${PYTHON_VERSION_STRING})
-    message(STATUS "Found PYTHON_EXECUTABLE=${PYTHON_EXECUTABLE}")
+    message(STATUS "Found Python_EXECUTABLE=${Python_EXECUTABLE}")
     message(STATUS "Python version ${PYTHON_VERSION_STRING}")
   endif()
-  
-  find_package(PythonLibs 3)
-  
+
   if (PYTHONLIBS_FOUND)
```
KrisThielemans, line 107: Python_*, and different conditional

SuperBuild.cmake
```diff
@@ -125,14 +123,16 @@ else()
...
     set_property(CACHE PYTHON_STRATEGY PROPERTY STRINGS PYTHONPATH SETUP_PY CONDA)
   endif()
 
   # set PYTHONLIBS_CMAKE_ARGS to be used in the ExternalProject_add calls
   # note: Find_package(PythonLibs) takes PYTHON_INCLUDE_DIR and PYTHON_LIBRARY as input
-  set (PYTHONLIBS_CMAKE_ARGS -DPYTHON_EXECUTABLE:FILEPATH=${PYTHON_EXECUTABLE})
+  set (PYTHONLIBS_CMAKE_ARGS
+       -DPython_EXECUTABLE:FILEPATH=${Python_EXECUTABLE}
+       -DPython3_EXECUTABLE:FILEPATH=${Python_EXECUTABLE})
   if (EXISTS "${PYTHON_INCLUDE_DIR}")
```
KrisThielemans, line 136: Python, not PYTHON

KrisThielemans (changes requested):

SuperBuild/External_NiftyPET.cmake
```diff
@@ -68,7 +68,8 @@ if(NOT ( DEFINED "USE_SYSTEM_${externalProjName}" AND "${USE_SYSTEM_${externalPr
       -DCMAKE_PREFIX_PATH=${SUPERBUILD_INSTALL_DIR}
       -DCMAKE_LIBRARY_PATH=${SUPERBUILD_INSTALL_DIR}/lib
       -DCMAKE_INSTALL_PREFIX=${${proj}_INSTALL_DIR}
-      -DPYTHON_EXECUTABLE=${NiftyPET_PYTHON_EXECUTABLE}
+      -DPython_EXECUTABLE=${NiftyPET_PYTHON_EXECUTABLE}
+      -DPython3_Executable=${NiftyPET_PYTHON_EXECUTABLE}
```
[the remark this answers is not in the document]
casperdcl: Any reason why? Current NiftyPET is Python3 only.

KrisThielemans:
There seem to be 2 different Python versions used in the CIL install, resulting in the `cil` not being found at import. There are 2 causes for this.

For the SB, CMake finds a surprising version (which is then passed on)
```
Found Python: /opt/hostedtoolcache/Python/3.10.12/x64/bin/python3.10 (found version "3.10.12") found components: Interpreter Development Development.Module Development.Embed
Found Python_EXECUTABLE=/opt/hostedtoolcache/Python/3.10.12/x64/bin/python3.10
```
The CMake strategy for selecting a python version is still far too complicated/error-prone. I suggest therefore to explicitly set
`Python_EXECUTABLE=`which python` in the Action.

In addition, it looks like the CIL version that is being built (I only checked with `DEVEL_BUILD=OFF`) uses old-style:
```
-- Found PythonLibs: /usr/lib/x86_64-linux-gnu/libpython3.10.so (found version "3.10.12")
-- Found PYTHON_EXECUTABLE=/home/runner/virtualenv/bin/python
```
I think therefore we need to add `PYTHON_EXECUTABLE` to the `PYTHONLIBS_CMAKE_ARGS` variable. That's safest in any case.

VirtualBox/scripts/UPDATE_functions.sh
```diff
@@ -185,7 +185,8 @@ SuperBuild_install(){
         -DDEVEL_BUILD="$DEVEL_BUILD" \
         -DNIFTYREG_USE_CUDA=OFF\
         -DBUILD_CIL=ON\
-        -DPYTHON_EXECUTABLE="$PYTHON_EXECUTABLE"\
+        -DPython_EXECUTABLE="$PYTHON_EXECUTABLE"\
+        -DPython3_EXECUTABLE="$PYTHON_EXECUTABLE"\
```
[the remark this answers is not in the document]
KrisThielemans: oh, right. This is on the VM. Sorry. Let's leave this in then.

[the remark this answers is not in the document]
KrisThielemans: Certainly as on GHA, CMake found the wrong python...

SuperBuild/External_NiftyPET.cmake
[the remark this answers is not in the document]
KrisThielemans: We currently only use NiftyPET via STIR, but that all happens on the C++ front. NiftyPET3 no longer has the appropriate libraries, and I have no energy for it.

See https://github.com/UCL/STIR/issues/1294 to jog your mind :-)

KrisThielemans:
```
ModuleNotFoundError: No module named 'Cython'
```
when building TomoPhantom. Not sure why this happens now? Wrong python used? TBH, We shouldn't really build TomoPhantom as part of our CI I think. @paskino ?

```
Run actions/upload-artifact@v4
  with:
    name: build_log_files-ubuntu-22.04-gcc9-Release-DEVEL=OFF
...
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```
I guess we have a problem with the naming somewhere.

casperdcl:
`TomoPhantom` `v2.0.0` (and also the slightly more recent `v3.0`) doesn't specify build deps. Its `setup.py` needs `Cython` & `numpy`. Adding them to the build step works.

Current test failures due to lack of CIL deps.

[CIL:scripts/requirements-test.yml](https://github.com/TomographicImaging/CIL/blob/master/scripts/requirements-test.yml) are not included in [CIL:pyproject.toml](https://github.com/TomographicImaging/CIL/blob/master/pyproject.toml) and I think they should be.

SuperBuild.cmake
```diff
@@ -125,24 +123,23 @@ else()
...
-  if (EXISTS "${PYTHON_INCLUDE_DIR}")
-    set (PYTHONLIBS_CMAKE_ARGS ${PYTHONLIBS_CMAKE_ARGS}
-      -DPYTHON_INCLUDE_DIR:PATH=${PYTHON_INCLUDE_DIR})
-  endif()
-  if (EXISTS "${PYTHON_LIBRARY}")
+  set (PYTHONLIBS_CMAKE_ARGS
+    -DPython_EXECUTABLE:FILEPATH=${Python_EXECUTABLE}
+    -DPython3_EXECUTABLE:FILEPATH=${Python_EXECUTABLE}
+    -DPYTHON_EXECUTABLE:FILEPATH=${Python_EXECUTABLE})
+  if (Python_Development_FOUND)
```
KrisThielemans, line 136: we require it, so no extra `if`

SuperBuild.cmake
```diff
@@ -125,24 +123,23 @@ else()
...
     set (PYTHONLIBS_CMAKE_ARGS ${PYTHONLIBS_CMAKE_ARGS}
-      -DPYTHON_LIBRARY:FILEPATH=${PYTHON_LIBRARY})
+      -DPython_INCLUDE_DIRS=${Python_INCLUDE_DIRS}
+      -DPython_LIBRARIES=${Python_LIBRARIES}
+      -DPYTHON_INCLUDE_DIR=${Python_INCLUDE_DIRS}
+      -DPYTHON_LIBRARY=${Python_LIBRARIES})
```
KrisThielemans, line 141: I don't think we should be setting the `*DIRS` and `*LIBRARIES` variables, only the "singular" ones, as they are the input ones, while the "plural" variables are the output. (It could go wrong when there are multiple directories in the "plural" variables).

That means though that we should only set those if the input ones are set.

KrisThielemans:
> For the SB, CMake finds a surprising version (which is then passed on)
> 
> ```
> Found Python: /opt/hostedtoolcache/Python/3.10.12/x64/bin/python3.10 (found version "3.10.12") found components: Interpreter Development Development.Module Development.Embed
> Found Python_EXECUTABLE=/opt/hostedtoolcache/Python/3.10.12/x64/bin/python3.10
> ```
> 
> The CMake strategy for selecting a python version is still far too complicated/error-prone. I suggest therefore to explicitly set `Python_EXECUTABLE=${which python}` in the Action.

This is still the case, hence the failures.

casperdcl:
Ah I finally understood; you *don't* want to use the system Python.

TL;DR we need to unset the cmake helpers defined by https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#environment-variables

KrisThielemans:
> Ah I finally understood; you _don't_ want to use the system Python.
> 
> TL;DR we need to unset [the cmake helpers defined by `actions/setup-python`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#environment-variables).

pretty confusing... I don't care which python is being used here, but obviously we should always use the same. As we seem to be using `setup-python` with `pip`, and then create a venv, we should be using that one. I guess that's what you did now, so ... great.

Sadly, some zenodo download issues ATM.

KrisThielemans (approved):

KrisThielemans (commented):
oh. still need to update CHANGES.md!

KrisThielemans:
Success! Thanks @casperdcl . however, can you now have a different PR to update CHANGES.md, saying that we've switched to FindPython (maybe a link), and that they should use `Python_EXECUTABLE` or any other variables as indicated in the CMake doc? (Sorry, I shouldn't have approved the PR :-))

casperdcl:
Already updated the changelog; feel free to update further

KrisThielemans:
Sorry. Confused. Which changelog?

casperdcl:
https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/CHANGES.md

PR #937, opened 2025-02-11
Update build for ubuntu 22.04

paskino:
- fixes #935

Defers the installation of CIL run requirements which confuse CMake.

PR #943, opened 2025-03-11
GHA: use Ubuntu 22.04

paskino:
Updates the GHA to explicitly use Ubuntu 22.04. 

Adds the `zenodo_get` dependency in `requirements.txt`.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -76,9 +76,9 @@ jobs:
       working-directory: docker
       run: |
         sudo ./raw-ubuntu.sh
-        USE_UBUNTU_SYSTEM_LIBRARIES=1 sudo ./build_system-ubuntu.sh
+        USE_UBUNTU_SYSTEM_LIBRARIES=1 sudo -E ./build_system-ubuntu.sh
         sudo ./build_essential-ubuntu.sh
-        USE_UBUNTU_SYSTEM_LIBRARIES=1 sudo ./build_gadgetron-ubuntu.sh
+        USE_UBUNTU_SYSTEM_LIBRARIES=1 sudo -E ./build_gadgetron-ubuntu.sh
         sudo ./build_python-ubuntu.sh
         echo "python-version=$(python3 --version | awk '{print $2}')" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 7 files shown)


paskino:
Thanks for your comments @KrisThielemans . Indeed my aim was to split #937 in subtasks, but I forgot to account for the VM, for instance.

I was aiming at fixing docker with a second PR, keeping these separated.

As we use the same build scripts for docker, VM and straight Ubuntu systems (the VM is one of them), we need to be able to differentiate what to install and what not. 

I understand now that I should not have broken backward compatibility.

Probably, I should have added a variable that is `INSTALL_PREREQUISITES_VIA_CONDA`. Does it make sense?

KrisThielemans:
> Probably, I should have added a variable that is `INSTALL_PREREQUISITES_VIA_CONDA`. Does it make sense?

that's a good idea. We could default it to 0 now, and switch to 1 later. But maybe you want to do that in the 2nd PR anyway (in which case this PR would be really small). As you prefer.

.github/workflows/c-cpp.yml
```diff
@@ -76,9 +76,9 @@ jobs:
       working-directory: docker
       run: |
         sudo ./raw-ubuntu.sh
-        sudo ./build_system-ubuntu.sh
+        USE_UBUNTU_SYSTEM_LIBRARIES=1 sudo -E ./build_system-ubuntu.sh
```
KrisThielemans, line 79: still need to undo this

KrisThielemans (changes requested):

.github/workflows/c-cpp.yml
KrisThielemans:
```suggestion
        sudo ./build_system-ubuntu.sh
```

.github/workflows/c-cpp.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         sudo ./raw-ubuntu.sh
         sudo ./build_system-ubuntu.sh
         sudo ./build_essential-ubuntu.sh
-        sudo ./build_gadgetron-ubuntu.sh
+        USE_UBUNTU_SYSTEM_LIBRARIES=1 sudo -E ./build_gadgetron-ubuntu.sh
```
KrisThielemans, line 81:
```suggestion
        sudo ./build_gadgetron-ubuntu.sh
```

PR #956, opened 2025-06-23
Update VirtualBox/documentation/INSTALL.md with macOS(ARM64) instruction

ColomboMatte0:
Hi,
I’ve modified the INSTALL.md file to include the steps that worked for me to run the VM on a macOS system with Apple Silicon (ARM64).

These instructions cover:
	•	Installing dependencies such as Homebrew and QEMU
	•	Extracting and converting the VM image
	•	Launching the VM using qemu-system-x86_64 with port forwarding

On my laptop (M4 Pro), the VM GUI was extremely laggy and almost unusable. For this reason, I’ve included instructions to set up port forwarding and access the VM through my own browser instead.

I also attempted to use UTM, a GUI wrapper for QEMU on macOS, but was unable to get the VM image working there — which is why I opted for the more minimal QEMU CLI setup.

It would be helpful if someone else with a Mac could test these steps to confirm whether they work more broadly.

### DIFF
--- a/VirtualBox/INSTALL.md
+++ b/VirtualBox/INSTALL.md
@@ -3,6 +3,8 @@
 A pre-built VM can be downloaded from this [Zenodo page](https://zenodo.org/records/10887535). Here we explain
 how to install it, once downloaded. (If you want to build a new VM yourself, then check the [vagrant/README.md](vagrant/README.md) for instructions.)
 
+For macOS systems with Apple Silicon (ARM64), follow the dedicated section below.
+
 ## Initial download and installation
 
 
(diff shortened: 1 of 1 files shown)


KrisThielemans:
@paskino I think you're the only one available with an ARM Mac.

VirtualBox/INSTALL.md
```diff
@@ -32,6 +34,66 @@ Warning: this file can be ~4.9GB. (You can download to a USB stick or hard drive
...
+- Set a password for the Jupyter server:
+```bash
+jupyter notebook password
+```
+9. Launch the jupyter server with port forwarding:
+```bash
+jupyter lab --ip=0.0.0.0 --port=8888 --no-browser
+```
+10. Open the jupyterlab from outside the VM
+- open a browser (preferably not safari) and go to [http//localhost:8888](http//localhost:8888)
```
KrisThielemans, line 90:
```suggestion
- open a browser (preferably not safari) and go to [http://localhost:8888](http://localhost:8888)
```

VirtualBox/INSTALL.md
```diff
@@ -32,6 +34,66 @@ Warning: this file can be ~4.9GB. (You can download to a USB stick or hard drive
...
+  -display cocoa \
+  -vga std
+```
+- -m specifies the RAM size. Use about half of your laptop’s total RAM. Assigning too much RAM can slow down your host machine, while too little will slow down the virtual machine 
+- -smp sets the number of CPU cores to emulate (x86_64).
+- -net user,hostfwd=tcp::8888-:8888 exposes port 8888 of the VM to the host.
+
+
+8. Log in and set up JupyterLab:
+
```
KrisThielemans, line 77:
```suggestion
   > [!NOTE]
   > In our experience, the GUI performance through QEMU is quite laggy, so we recommend to either access JupyterLab through your browser (see below), or you can use `ssh -X`.**
```

Issue #959, opened 2025-06-26
failed to import CIL on one of the GitHub Actions jobs


KrisThielemans:
See https://github.com/SyneRBI/SIRF-SuperBuild/pull/958#issuecomment-3008197582

casperdcl:
Fixed by #963?

PR #965, opened 2025-10-23
Removed RocksDB dependency from Gadgetron

### LINKED ISSUE #940: Gadgetron no longer needs RocksDB
We can therefore get rid of it as dependency. (I suggest we leave it in as an eternal project)

KrisThielemans:
Fixes #940


KrisThielemans:
Normal jobs ok, 1 job failure due #959.

I'm going to merge.

PR #970, opened 2025-11-03
bumpy CCPi-Regularisation Toolkit to version compatible with numpy 2+

paskino:
Updated the CCPi-Regularisation-Toolkit version to v25.0.0 which is compatible with numpy 2.x

### DIFF
--- a/docker/requirements.yml
+++ b/docker/requirements.yml
@@ -34,7 +34,7 @@ dependencies:
   - tqdm              # cil
   - numba             # cil
   ## - tigre          # cil (GPU) # TODO: remove second # prefix once tigre is build for numpy>=2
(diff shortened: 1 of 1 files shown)


docker/requirements.yml
```diff
@@ -34,7 +34,7 @@ dependencies:
   - tqdm              # cil
   - numba             # cil
   ## - tigre          # cil (GPU) # TODO: remove second # prefix once tigre is build for numpy>=2
-  # - astra-toolbox 2 cuda* # cil (GPU)
+  # - astra-toolbox=2.1=cuda* # cil (GPU)
```
casperdcl: 2.1 might not work; best to revert this


```suggestion
  # - astra-toolbox 2 cuda* # cil (GPU)
```

casperdcl (approved):

PR #973, opened 2025-11-20
HDF5 fixes

KrisThielemans:
STIR and SIRF need CXX, ISMRMRD needs C

Fixes #972

PR #976, opened 2025-11-28
SIRF CMake improvements

KrisThielemans:
-  adding `SIRF_DISABLE_Gadgetron_toolboxes` makes it easier for the user to disable this if there are compilation problems
- prevent adding SIRF options and dependencies when `USE_SYSTEM_SIRF==ON`

(Note: sits on top of #973 and #975, although it's independent, so let's merge those first, or if not, we rebase this on current master).


KrisThielemans:
I expect the same docker and run-time failure as in #973 but no other.

PR #985, opened 2026-02-16
use startup hooks

casperdcl:
See also:
- https://github.com/SyneRBI/PETRIC-backend/pull/7
- https://github.com/jupyter/docker-stacks/issues/2386

TODO:
- [ ] partially revert https://github.com/SyneRBI/PETRIC-backend/pull/7 post-merge


KrisThielemans:
Looks like a good idea to me. I'll let you sort out what it fails :-)

PR #996, opened 2026-05-18
llvm-openmp to fix CIL runtime error, misc housekeeping

casperdcl:
Three unrelated items; can't be bothered to split into separate PRs:

1. **mamba error** (info below) CI: use `conda` instead of `mamba`
2. deps (housekeeping): move `ipywidgets` dependencies from `pip` to `conda` due to https://github.com/TomographicImaging/CIL/issues/1600 being resolved
3. deps (bugfix): use `llvm-openmp` (instead of nothing or `libgomp`) to fix CIL runtime error ([no idea why](https://github.com/SyneRBI/SIRF-SuperBuild/pull/996#issuecomment-4485736315))

## 1. mamba error

Comparing [-passing](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/25919844427/job/76493109354) to [+failing](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/25919844427/job/76493108948):

```diff
 Image: ubuntu-22.04
-Version: 20260413.88.1
+Version: 20260513.139.2
...
 Downgrade:
 - _openmp_mutex                                 5.1  1_gnu                     pkgs/main
 + _openmp_mutex                                 4.5  7_kmp_llvm                conda-forge
+- libarchive                                   3.8.7  hb3cce40_0               pkgs/main
++ libarchive                                   3.8.6  gpl_hc2c16d8_100         conda-forge
+- libmamba                                     2.6.1  hd28c85e_0               conda-forge
++ libmamba                                     2.5.0  hd28c85e_0               conda-forge
+- libmamba-spdlog                              2.6.1  hd8d719e_0               conda-forge
++ libmamba-spdlog                              2.5.0  h12fcf84_0               conda-forge
+- libmambapy                                   2.6.1  py313h842faf6_0          conda-forge
++ libmambapy                                   2.5.0  py313h4616538_0          conda-forge
+- mamba                                        2.6.1  h2a27def_0               conda-forge
++ mamba                                        2.5.0  h9835478_0               conda-forge
+- simdjson                                     4.6.4  hb700be7_0               conda-forge
++ simdjson                                     4.2.4  hb700be7_0               conda-forge
 Summary:
 Install: 277 packages
-Change: 3 packages
-Upgrade: 8 packages
-Downgrade: 1 packages
+Change: 2 packages
+Upgrade: 5 packages
+Downgrade: 7 packages
...
+/tmp/mambafylm1mdqvkn: line 2: /usr/share/miniconda/bin/mamba (deleted): No such file or directory
+/tmp/mambafylm1mdqvkn: line 3: syntax error near unexpected token `deleted'
+/tmp/mambafylm1mdqvkn: line 3: `mamba (deleted) activate "/usr/share/miniconda"'
```

So somehow `mamba` is being downgraded `2.6.1` -> `2.5.0` causing failures.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -75,7 +75,7 @@ jobs:
       name: install dependencies
       working-directory: docker
       run: |
-        mamba env update -n base -f requirements.yml
+        conda env update -n base -f requirements.yml
         echo "Python_ROOT_DIR=$CONDA_PREFIX" >> $GITHUB_ENV
         echo "Python3_ROOT_DIR=$CONDA_PREFIX" >> $GITHUB_ENV
         sudo ./raw-ubuntu.sh
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -208,6 +208,8 @@ if (NOT DISABLE_OpenMP)
   find_package(OpenMP)
   if (OPENMP_FOUND)
      message(STATUS "OpenMP found, support enabled. If you get compiler errors saying \"omp.h not found\", set the following CMake variables: OpenMP_CXX_INCLUDE_DIR, OpenMP_C_INCLUDE_DIR. For linking errors, check other \"OpenMP\*\" CMake variables as well.")
+     message(STATUS "OpenMP_CXX_INCLUDE_DIR: ${OpenMP_CXX_INCLUDE_DIR}")
+     message(STATUS "OpenMP_CXX_LIB_NAMES: ${OpenMP_CXX_LIB_NAMES}")
 
     mark_as_superbuild(ALL_PROJECTS VARS
       OpenMP_CXX_FLAGS:STRING OpenMP_C_FLAGS:STRING
--- a/docker/requirements.yml
+++ b/docker/requirements.yml
@@ -24,7 +24,7 @@ dependencies:
   - jupytext
   - tensorboard
   - tensorboardx
-  - libgomp # parallelproj
+  - llvm-openmp
   - jupyterlab_widgets
   - ipywidgets >=8
   - ipympl
--- a/docker/requirements.yml
+++ b/docker/requirements.yml
@@ -25,6 +25,9 @@ dependencies:
   - tensorboard
   - tensorboardx
   - libgomp # parallelproj
+  - jupyterlab_widgets
+  - ipywidgets >=8
+  - ipympl
   # - ccpi/label/dev::tigre # cil (GPU)
   # - astra-toolbox::astra-toolbox 2.* # cil (GPU)
   # - cuda-version 12.8.* # cil (GPU)
@@ -45,11 +48,4 @@ dependencies:
   - pip
   - pip:
     - git+https://github.com/ismrmrd/ismrmrd-python-tools.git@master#egg=ismrmrd-python-tools
-    # TODO: labextension @jupyter-widgets/jupyterlab-manager (jupyterlab_widgets ipywidgets ipympl)
-    # is broken because `conda-forge::cil` downgrades `ipywidgets<8` -> breaks `%matplotlib widgets`.
-    # - temp fix: install upgraded versions via pip (not conda)
-    # - proper fix: remove after https://github.com/TomographicImaging/CIL/issues/1600
-    - jupyterlab_widgets
-    - ipywidgets>=8
-    - ipympl
     - unittest-parametrize # cil


KrisThielemans:
Personally, I think openmp needs to be the same as whatever is installed with CIL. Not sure if conda will sort this out. Fingers crossed.

KrisThielemans:
seems to have worked anyway :-). 

Of course, it might break when using conda-forge version of STIR, as that uses `libgomp` on Linux. https://github.com/conda-forge/stir-feedstock/blob/eb4e628560e381dfaf3f497519fe206f867f5049/recipe/recipe.yaml#L46-L49

Ideally read https://conda-forge.org/docs/maintainer/knowledge_base/#openmp before merging, although that doesn't talk about Intel OpenMP. There's also https://stackoverflow.com/questions/76653505/intelomp-and-llvm-omp-colliding

casperdcl:
CIL is built with [SuperBuild/External_CIL.cmake](https://github.com/SyneRBI/SIRF-SuperBuild/blob/master/SuperBuild/External_CIL.cmake) which eventually calls [CIL@numpy-2:src/Core/CMakeLists.txt](https://github.com/TomographicImaging/CIL/blob/numpy-2/src/Core/CMakeLists.txt) which is effectively:
```cmake
project(cil LANGUAGES CXX)
find_package(OpenMP REQUIRED)
add_definitions(${OpenMP_CXX_FLAGS})
add_library(cilacc SHARED ${cilacc_SOURCES})
target_link_libraries(cilacc OpenMP::OpenMP_CXX)
```
... do you see anything incompatible?

KrisThielemans:
I see that when installing `llvm-openmp` explicitly, everything worked, while when removing `libgomp`, which I think is safer, it fails again. This seems to be because CMake [found](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/26063590199/job/76629266128#step:14:376) the system OpenMP (which is `libomp`).

> CIL is built with ...
> ... do you see anything incompatible?

Of course not. The issue is which openMP library is found (and linked with). The [openmp run-time error was/is](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/26036172348/job/76534805902#step:14:3430)
```
OMP: Error #15: Initializing libomp.so, but found libiomp5.so already initialized.
5: OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://openmp.llvm.org/
```
(BTW, I cannot immediately find anything relevant on the link suggested at the end of that message)

We saw that (for whatever reason) `conda install .... cil` [installs](https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/26036172348/job/76534805902#step:5:38) `intel-openmp`. We were installing `libgomp` (as recommended by conda-forge on Linux). Presumably, (bits from) CIL were linked against the intel version, and SIRF (and STIR) with the GNU version.

Presumably, when forcing `llvm-openmp`, there's no run-time conflict with `intel-openmp` and it works. I'm just afraid this will occur again when using conda-forge STIR. When not explicitly install an openmp library, in the hope of picking up whatever was there already, CMake doesn't find the intel version (maybe `intel-openmp` doesn't include header files or whatever). 

Anyway, I suggest we go ahead with `llvm-openmp` and see later what happens when using conda for more dependencies. (Maybe rename the title of this PR such that it'll be easier to find).

KrisThielemans:
What a mess though.

casperdcl:
Kris we're not doing `conda install cil`.
1. we `apt`/`conda` install **dependencies**
2. we SuperBuild (i.e. `cmake --install` -> `pip install --no-deps`) lots of things (including CIL) none of which install omp libs
3. all the SuperBuild's `External_*.cmake` should eventually call `find_package(OpenMP)` and thus all use the same libomp, right?

For point (1) dependencies:
a) cil complains at runtime if this includes `conda-forge::libgomp` (which seems to include `intel-openmp`)
b) cil complains at runtime if this doesn't include any conda omp lib
c) cil is fine with `conda-forge::llvm-openmp`

> Presumably, (bits from) CIL were linked against the intel version, and SIRF (and STIR) with the GNU version

Why, if all use `find_package(OpenMP)` and `target_link_libraries(... OpenMP::OpenMP_CXX)`?

Furthermore, the docker build is fine with any strategy (no dep, `libgomp`, `llvm-openmp` etc).

KrisThielemans:
Clearly, I don't know anymore how/when/where CIL is built.

> For point (1) dependencies:
> a) cil complains at runtime if this includes `conda-forge::libgomp` (which seems to include `intel-openmp`) 

`conda-forge::libgomp` doesn't include `intel-openmp` on its own. I guess the latter is being dragged in by something else (such as pytorch, MKL, TBI, whatever)

> 
> > Presumably, (bits from) CIL were linked against the intel version, and SIRF (and STIR) with the GNU version
> 
> Why, if all use `find_package(OpenMP)` and `target_link_libraries(... OpenMP::OpenMP_CXX)`?

If they are all built by the SuperBuild, then this is indeed pretty weird. I guess can only be found out by diving in, and spending 2 days or more on it (which we don't want to do!)

I have no answers here! Sorry.

Let's just merge and pray it doesn't break tomorrow.

Issue #470, opened 2021-01-14
SWIG on Windows installed in confusing directory name


KrisThielemans:
SWIG is installed in this directory
https://github.com/SyneRBI/SIRF-SuperBuild/blob/199921c3305a1acd1c95d8de4ebfdcdd00be286b/SuperBuild/External_SWIG.cmake#L39
even if it is a different version. We should either remove any version info from the name (easiest), or fix it somehow.

KrisThielemans:
it's a bit surprising actually that the version info disappears as the downloaded zip file extracts in `swigwin-4.0.2` for instance.

PR #999, opened 2026-05-27
STIR 6.4.0

KrisThielemans:


KrisThielemans:
https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/26536045198/job/78165075196?pr=999#step:10:17186
```
                 from /home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/builds/STIR/build/src/swig/CMakeFiles/_stir.dir/stirPYTHON_wrap.cxx:4033:
/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/sources/STIR/src/include/stir/assign.h: In instantiation of ‘std::enable_if_t<stir::has_iterator_and_no_full_iterator<T>::value> stir::assign(T&, const T2&) [with T = swig::SwigPySequence_Cont<stir::DetectionPositionPair<> >; T2 = std::vector<stir::DetectionPositionPair<> >*; std::enable_if_t<stir::has_iterator_and_no_full_iterator<T>::value> = void]’:
...
/home/runner/work/SIRF-SuperBuild/SIRF-SuperBuild/build/sources/STIR/src/include/stir/assign.h:70:3: error: cannot bind non-const lvalue reference of type ‘swig::SwigPySequence_Ref<stir::DetectionPositionPair<> >&’ to an rvalue of type ‘swig::SwigPySequence_InputIterator<stir::DetectionPositionPair<>, swig::SwigPySequence_Ref<stir::DetectionPositionPair<> > >::reference’ {aka ‘swig::SwigPySequence_Ref<stir::DetectionPositionPair<> >’}
   70 |   for (auto& i : v)
```
So, this seems to be a STIR problem, which for whatever reason transpires on the SB, while not in STIR-CI nor conda-forge (multiple compilers/OSs). Sigh

KrisThielemans:
i see that these are all with `-swig`, so APT swig, which is
https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/26536045198/job/78165075150?pr=999#step:9:142
```
Found SWIG: /usr/bin/swig4.0 (found suitable version "4.0.2", minimum required is "3")
```
the +swig job works ok, which uses
https://github.com/SyneRBI/SIRF-SuperBuild/blob/80f851cedb86cf0a545c0b5da5be6b0dfcc65a9d/version_config.cmake#L84-L90

So, it seems STIR now needs quite recent SWIG. I suppose no alternative than building SWIG (or getting it from conda-forge)

docker/requirements.yml
```diff
@@ -10,6 +10,7 @@ channels:
 dependencies:
   - setuptools
   - wheel
+  - swig >=4.1
```
casperdcl: Based on https://www.swig.org/Release/RELEASENOTES I'm unsure which is most appropriate as a minimum betwixt [`4.1`, ..., `4.4`].

The CI is currently passing with `conda-forge::swig=4.4.1` and/or `http://downloads.sourceforge.net/swig/swig-4.4.0.tar.gz`.

Will push a debug commit to test all of the above versions here.

KrisThielemans: yes, it's not clear when this happened. It's worth testing. I've created https://github.com/UCL/STIR/issues/1721 showing that 4.2 is fine, so would it be easy to test 4.1 for you?

KrisThielemans: oh, you're doing it already :-)

KrisThielemans: Lots of time-outs in the previous run, but re-running hits the problem that the MD5 doesn't match
https://github.com/SyneRBI/SIRF-SuperBuild/blob/80f851cedb86cf0a545c0b5da5be6b0dfcc65a9d/version_config.cmake#L89

easiest (?) might be to install it via conda and use "system" SWIG. However, we only need to check 4.1, so could just try that one and adjust the MD5.

casperdcl: passing:

- `downloads.sourceforge.net/swig>=4.2`
- `conda-forge::swig>=4.4` (failing to `conda install` with anything less (!))

vis. https://github.com/SyneRBI/SIRF-SuperBuild/actions/runs/26760693898?pr=999 from https://github.com/SyneRBI/SIRF-SuperBuild/pull/999/commits/131294b72a732fe9034f5b2dbe08f1e8d9ad48bb

casperdcl: I'm suggesting >=4.2 everywhere; and let conda resolve to >=4.4 on its own

KrisThielemans: ok. let's do that. no use wasting more time on it! Thanks a lot.

KrisThielemans: There's also #470, but I think that's unrelated, so feel free to merge this.

PR #1003, opened 2026-06-05
CMake set RPATH, CI/docker/VM: Ubuntu 24.04, update ISMRMD/Gadgetron

KrisThielemans:
Fixes #934

- [ ] maybe try 26.04?
- [ ] related: https://github.com/SyneRBI/SIRF/issues/1359

PR #1013, opened 2026-06-24
docker: local cache, devel build with full build env

casperdcl:
- [x] docker devel images keep the full build environment (cmake & deps) via new `target: sirf-dev`
  + [x] also fixes `/opt` permissions to be `users` group owned
- [x] use docker [local cache mounts](https://docs.docker.com/reference/dockerfile/#run---mounttypecache) for `apt` & `ccache`
- [x] CI cache: deprecate `docker/devel/.ccache` in favour of https://github.com/reproducible-containers/buildkit-cache-dance
  - [x] fix buildkit support https://docs.docker.com/compose/how-tos/dependent-images/#use-another-services-image-as-the-base-image
- [x] update docs

Related: #1003 etc. Helps building & debugging locally:

```sh
./docker/compose.sh -bdc -- --build-arg STIR_DISABLE_HDF5=ON
```

builds in ~100s on my machine (using `--build-arg RUN_CTEST=0`).


Dockerfile
```diff
@@ -111,6 +108,34 @@ RUN echo 'test -z "$OMP_NUM_THREADS" && export OMP_NUM_THREADS=$(python -c "impo
...
+#ENV USE_HTTPS="yes"
+# gadgetron
+EXPOSE 9002
+ENV GADGETRON_RELAY_HOST="0.0.0.0"
+COPY --link docker/15source-sirf.sh docker/25clone-work.sh docker/35gadgetron.sh /usr/local/bin/before-notebook.d/
+ENV NOTEBOOK_ARGS=--PasswordIdentityProvider.hashed_password='sha1:cbf03843d2bb:8729d2fbec60cacf6485758752789cd9989e756c'
+
+# copy ccache from cache mount
+RUN --mount=type=cache,target=/opt/ccache cp -a /opt/ccache /ccache && fix-permissions /ccache
+RUN mv /ccache /opt/
```
KrisThielemans: first copy, then `mv`?

KrisThielemans (commented):
Looks fine, but it does go beyond what I know :-)

KrisThielemans:
What about permissions? They're still wrong do anything useful as a developer with this image (although it does help of course, so we should merge this)

Dockerfile
casperdcl: just trying to get the `--mount`, well, unmounted and baked into an image layer:

- mount /opt/ccache
- copy /opt/ccache -> /ccache
- unmount /opt/ccache
- mv /cache -> /opt/cache

casperdcl:
good point - fixed the perms now :)

KrisThielemans:
Great. (I can't see where, but never mind!).

Will need some changes to `docker/README.md` and `CHANGES.md`

KrisThielemans (commented):
I still don't understand all of it, but looks good.

Some docs in README and CHANGES, please :-)

PR #1016, opened 2026-06-29
CI: hdf5 logic

### LINKED ISSUE #980: docker: STIR_DISABLE_HDF5=OFF
Would require fixing linker errors

casperdcl:
Really fixes #980 (we fixed docker in #1014).
Related: #997.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -146,7 +146,7 @@ jobs:
         set -ex
         export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
         export superbuild=`pwd`
-        BUILD_FLAGS="-DCMAKE_INSTALL_PREFIX=~/install -DPYVER=3 -DUSE_SYSTEM_RocksDB:BOOL=OFF -DBUILD_TESTING_Gadgetron:BOOL=ON -DSTIR_DISABLE_HDF5:BOOL=ON -DUSE_ITK:BOOL=ON -DBUILD_CIL:BOOL=ON -DBUILD_siemens_to_ismrmrd:BOOL=ON"
+        BUILD_FLAGS="-DCMAKE_INSTALL_PREFIX=~/install -DPYVER=3 -DUSE_SYSTEM_RocksDB:BOOL=OFF -DBUILD_TESTING_Gadgetron:BOOL=ON -DUSE_ITK:BOOL=ON -DBUILD_CIL:BOOL=ON -DBUILD_siemens_to_ismrmrd:BOOL=ON"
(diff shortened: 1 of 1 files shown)


casperdcl:
*cough*... should we drop support for ubuntu 22.04 system HDF5 @KrisThielemans?

KrisThielemans:
Right. I guess I was a bit over-optimistic...

Error is when link SIRF utilities (all the rest seems fine):
```
/usr/bin/ld: /usr/share/miniconda/lib/libhdf5_cpp.so.320.0.0: undefined reference to `__cxa_call_terminate@CXXABI_1.3.15'
```
This is caused by conda libs being compiled with g++14 (or later?), while we're using g++9 or 11 here. Apparently, gcc14 added `__cxa_call_terminate` to the C++ ABI.

Of course, conda recommends using its own pre-packaged compiler to avoid this stuff, as I do at https://github.com/SyneRBI/SIRF/wiki/Building-SIRF-and-CIL-with-conda.

So,  I guess we have the following options (sorted in order of my preference)

1. `conda install cxx-compiler` and test with that (needs some change in the CI logic.)
2. Keep `SIRF_DISABLE_HDF5=ON` on Ubuntu 22.04
3. Don't test Ubuntu 22.04
4. Install g++14 on Ubuntu 22.04 ourselves, which is likely painful (self-compile? No way). It's also a losing battle.

Of course, on the Wiki page, I had `cxx-compiler=1.5.2`, which [is g++11](https://github.com/conda-forge/compilers-feedstock/pull/52), which probably goes back to old versions of everything anyway, and this problem would disappear even when using g++9! However, I'm optimistic we can drop the `cxx-compiler` limit (or at least make it more recent), as otherwise the Ubuntu 22.04 builds would have failed.

So, I vote for
- use option 2 above now (i.e. close this PR. done!)
- open another PR for option 1. This could be done in a few ways
  - add a new workflow using conda builds as much as possible (i.e. avoid installing pretty much anything via APT)
  - either 
    - update `requirements.yml` (but that'll affect the docker image), 
    - create a new one for now (not pretty, I know)
    - add a script somewhere that more or less follows what's on the wiki (but cutting old work-arounds), and test that.

I think this would be good as a step to finally tackle distributing via conda.

Aside from this, we should add something in https://github.com/SyneRBI/SIRF-SuperBuild/#faqs along the lines of
```
## linking errors like `undefined reference to `__cxa_call_terminate@CXXABI_1.3.15`

You might be mixing recent libraries and old compilers, e.g. recent conda-forge libraries and g++ < 14. When using conda, we recommend using conda's `cxx-compiler`. Alternatively, use a more recent compiler.
```

KrisThielemans:
Let's add an `if ubuntu < 24.04` or maybe check `gcc` version.

PR #1023, opened 2026-07-23
Pugixml shared and static, and v1.16

### LINKED ISSUE #1022: ISMRMRD can fail to build due to missing pugixml::shared target
More info is on https://github.com/ismrmrd/ismrmrd/issues/271

Easiest solution seems to be to build both static and shared versions of pugixml

KrisThielemans:
Fixes #1022 

also upgrade to 1.16 as it fixed some issues with more recent compilers etc.


KrisThielemans:
This PR is working fine on our university cluster (while without it the build fails due to #1022). Note that it's actually not tested in CI as we have
https://github.com/SyneRBI/SIRF-SuperBuild/blob/cec1da6efdacf501621bc60453ed80db9997ca9f/SuperBuild.cmake#L187
We probably should add a `+pugixml` job to our matrix.

PR #1026, opened 2026-08-05
devcontainer-ish

casperdcl:
Add devcontainer config for local VSCode development (`cmake -DDEVEL_BUILD=ON -DSTIR_BUILD_EXECUTABLES=ON -DRUN_CTEST=0 --build /opt/SIRF-SuperBuild` files still present, e.g. SIRF is cloned to `/opt/SIRF-SuperBuild/sources/SIRF`).

- docker: default `STIR_BUILD_SHARED_LIBS=ON`
- `.devcontainer.json` (default: CPU, need to manually modify to use GPU)
  + includes some helpful extensions (Python, Jupyter, CMake, etc.)
  + uses a few overrides in `docker/docker-compose.devcontainter.yml`
    * (to fix permissions for mounted volumes, modify `group_add` here)
  + makes `/opt/SIRF-SuperBuild` writable & persistent
    * to reset e.g. after updating the image: `docker compose down; docker volume rm sirf-superbuild_persistent`
  + also mounts `/opt/.devcontainer.json` for easier editing

When launched, you can run e.g. `cmake --build . --target SIRF` to (re)build after modifying sources.

Note: this is NOT the same approach as https://github.com/SyneRBI/SIRF-Exercises/tree/master/.devcontainer

> [!TIP]
> When opening the repo, VSCode will notice the config by itself and prompt to re-open in a devcontainer.

- [x] someone else test that this also works for them
- [x] test `git push` within the container: https://code.visualstudio.com/remote/advancedcontainers/sharing-git-credentials

## docs

Probably unnecessary to add these docs in the wiki given that the inline comments in `.devcontainer.json` already explain all this...

Intended for local VSCode development - CMake build flags (`DEVEL_BUILD=ON`, `STIR_BUILD_SHARED_LIBS=ON`, `STIR_BUILD_EXECUTABLES=ON`, `RUN_CTEST=0`), VSCode extensions (Python, Jupyter, CMake, etc.)

1. When opening SIRF-SuperBuild in VSCode, you will be prompted to `Reopen in Container` (because a [`.devcontainer.json`](https://aka.ms/devcontainer.json) config file was detected). Alternatively, use the Command Palette to `Rebuild and Reopen in Container`. :warning: BEFORE opening in a Dev Container:
   - for GPU support, (un)comment the relevant lines in `.devcontainer.json`
   - for sharing data, add extra volume mounts AND modify `group_add` in `docker/docker-compose.devcontainter.yml`
2. A docker image will be (re)built
3. If a docker volume called `sirf-superbuild_persistent` doesn't exists, it will be created AND populated with the image's `/opt/SIRF-SuperBuild`
4. A container is launched with the volume mounted on top of (overriding) the image's `/opt/SIRF-SuperBuild`

> [!NOTE]
> To force re-create the persistent volume (i.e. to discard your workspace after an image update), then
> 1. close VSCode
> 2. in the SIRF-SuperBuild directory on the host, run `docker compose down; docker volume rm sirf-superbuild_persistent`

When launched, you can run e.g. `cmake --build . --target SIRF` to (re)build after modifying sources.

Things like [`git push` should "just work" even inside the container](https://code.visualstudio.com/remote/advancedcontainers/sharing-git-credentials).


.devcontainer.json
```diff
@@ -0,0 +1,37 @@
+/// format: https://aka.ms/devcontainer.json
+/// config: https://github.com/devcontainers/templates/tree/main/src/docker-existing-docker-compose
+/// example: https://github.com/devcontainers-community/templates-jupyter-datascience-notebooks/tree/main/.devcontainer
+{
+	/// CPU
+	// "initializeCommand": "./docker/compose.sh -bdc -- --build-arg RUN_CTEST=0",
```
casperdcl:
```suggestion
	"initializeCommand": "./docker/compose.sh -bdc -- --build-arg RUN_CTEST=0",
```

.devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	// "dockerComposeFile": ["docker-compose.yml", "docker/docker-compose.devel.yml", "docker/docker-compose.devcontainer.yml"],
```
casperdcl:
```suggestion
```

.devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	"image": "synerbi/sirf:devel",
+	/// GPU
+	// "initializeCommand": "./docker/compose.sh -bdg -- --build-arg RUN_CTEST=0",
+	// "dockerComposeFile": ["docker-compose.yml", "docker/docker-compose.devel.yml", "docker/docker-compose.gpu.yml", "docker/docker-compose.gpu-devel.yml", "docker/docker-compose.devcontainer.yml"],
+	// "image": "synerbi/sirf:devel-gpu",
+	// "hostRequirements": {"gpu": true},
```
casperdcl: probably worth splitting CPU/GPU as per https://github.com/SyneRBI/SIRF-Exercises/tree/master/.devcontainer

```suggestion
```

.devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	"name": "SyneRBI SIRF",
+	// "service": "sirf", "runServices": ["sirf"],
```
casperdcl:
```suggestion
```

paskino (approved):
We checked together with @casperdcl and @evgueni-ovtchinnikov 
We tested and will approve with changes as per comments:

- split configuration for CPU/GPU in 2 files
- uncomment `initializeCommand` line

.devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	/// https://aka.ms/dev-containers-non-root
+	//"runArgs": ["--user=1000:${localEnv:GROUPS}", "--group-add=users"],  // NOTE: you likely need to hardcode replace `${localEnv:GROUPS}` with your local group ID to fix workspace access permissions
+	"runArgs": ["--user=1000:1000", "--group-add=users"],
+	"overrideCommand": false, // use docker-stacks default
```
casperdcl: needs testing, but this might minimise confusion between host & container:

```suggestion
	"workspaceMount": "source=${localWorkspaceFolder}/.git,target=/opt/SIRF-SuperBuild/.git,type=bind,consistency=cached",
	"workspaceFolder": "/opt/SIRF-SuperBuild",
	"overrideCommand": false, // use docker-stacks default
```

.devcontainer.json
```diff
@@ -0,0 +1,37 @@
...
+	/// GPU
+	// "initializeCommand": "./docker/compose.sh -bdg -- --build-arg RUN_CTEST=0",
+	// "dockerComposeFile": ["docker-compose.yml", "docker/docker-compose.devel.yml", "docker/docker-compose.gpu.yml", "docker/docker-compose.gpu-devel.yml", "docker/docker-compose.devcontainer.yml"],
+	// "image": "synerbi/sirf:devel-gpu",
+	// "hostRequirements": {"gpu": true},
+	"name": "SyneRBI SIRF",
+	// "service": "sirf", "runServices": ["sirf"],
+	/// https://aka.ms/dev-containers-non-root
+	//"runArgs": ["--user=1000:${localEnv:GROUPS}", "--group-add=users"],  // NOTE: you likely need to hardcode replace `${localEnv:GROUPS}` with your local group ID to fix workspace access permissions
+	"runArgs": ["--user=1000:1000", "--group-add=users"],
```
casperdcl:
```suggestion
	"runArgs": [
		"--user=1000:${localEnv:GROUPS}", // you may have to hardcode replace `${localEnv:GROUPS}` with your local group ID to fix workspace access permissions
		"--group-add=users"],
```

KrisThielemans:
Good idea. I have no time to look into this, I'm afraid, so I won't review. Just go ahead without me.

docker-compose.yml
casperdcl, line 1: VSCode annoyingly insists on pulling `BASE_IMAGE` without first inspecting `additional_contexts`. Thus `BASE_IMAGE` needs to actually exist at DockerHub. Used `hello-world` as 'tis the smallest image afaik (few kB).

.devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+		"ms-vscode.cpptools",
+		"ms-python.isort",
+		"tamasfe.even-better-toml",
+		"GitHub.vscode-github-actions"
+	],
+	"settings": {
+		"python.defaultInterpreterPath": "/opt/conda/bin/python",
+		"C_Cpp.default.includePath": [
+			"/opt/SIRF-SuperBuild/sources/**/include",
+			"/opt/SIRF-SuperBuild/INSTALL/include/STIR-6.4"]}}},
```
KrisThielemans, line 37: will need to be adapted

KrisThielemans (changes requested):

.devcontainer.json
casperdcl: 1. could do `include/**` for now (since it's lower priority than `sources/**/include` anyway) to pick up any generated headers
2. instead, ideally fix VSCode CMake extension and use it as the intellisense provider (to use the existing cache, vis. https://github.com/microsoft/vscode-cmake-tools/issues/1370#issuecomment-804430926, https://github.com/microsoft/vscode-cmake-tools/issues/575#issuecomment-437799225, https://github.com/microsoft/vscode-cmake-tools/issues/4293, https://github.com/microsoft/vscode-cmake-tools/issues/1373)

.gitignore
```diff
@@ -1,16 +1,9 @@
...
-Gadgetron/
-SWIG/
-Armadillo/
-googletest/
+/.ninja_log
+/CTestTestfile.cmake
+/build.ninja
+/builds/
+/sources/
+/INSTALL/
```
casperdcl, line 6: just in case a user puts them in a subdir...

```suggestion
.ninja_log
CTestTestfile.cmake
build.ninja
builds/
sources/
INSTALL/
```

docker/docker-compose.devcontainer.yml
```diff
@@ -0,0 +1,8 @@
+services:
+  sirf-build: {build: {args: {RUN_CTEST: 0}}}
+  sirf:
```
casperdcl, line 3: should we add STIR EXEs @KrisThielemans?

- to the devcontainer

```suggestion
  sirf:
    args:
      EXTRA_BUILD_FLAGS: >
        -DDEVEL_BUILD:BOOL=ON
        -DSTIR_BUILD_EXECUTABLES:BOOL=ON
```

- or `Dockerfile` (all docker images)?
- or `version_config.cmake` (all change default)?

Alternatively we can leave as-is since this works within a few seconds in the devcontainer:

```sh
cmake . -DSTIR_BUILD_EXECUTABLES:BOOL=ON
cmake --build . --target STIR
stir_math ...
```

KrisThielemans: a few seconds? really? I find that the linking takes a fair amount of time.

On the other hand, if you want to experiment (!), STIR CMake option `BUILD_SHARED_LIBS` should now work, which would make the linking a lot faster (and keep the devcontainer roughly the same size). I wouldn't yet do this outside of Docker, as we don't do proper versioning of the STIR shared libs yet. (PR welcome :-) )

casperdcl: ok maybe a minute? `time` results (28-core server):


step | time/[m]
--|--
reconfigure | 0:02
rebuild STIR (including re-linking `parallelproj`, `ITK`, `STIR`) | 1:35
rebuild STIR (i.e. just re-linking again) | 0:19

It does incease the size of `INSTALL/bin` by 730MB so I see why it's best not to include in distributions 😅

I suppose the shared libs will solve both time and size issues.

KrisThielemans:
How are we doing with this PR? Would be nice to merge it soon, such that I can use it in a demo at UC Davis in ~3 weeks...

docker/docker-compose.devcontainer.yml
casperdcl: ok so with `STIR_BUILD_SHARED_LIBS=ON`, using `STIR_BUILD_EXECUTABLES=ON` only increases `INSTALL/bin` by ~50MB. Think it's best to do by default (at least in devcontainers if not all docker images in general).

casperdcl:
Tested & merged. Failing job is unrelated (due to #1030).

docker/docker-compose.devcontainer.yml
KrisThielemans: Not sure if you implemented this, but if the tests work, I think it's a great idea to have both ON on docker and devcontainer. Sadly, not yet for self-install (needs versioned libraries)

casperdcl: follow-up: #1033

